PR #21, opened 2021-03-08
Show a message on cookie accept/reject

apricot13:
@evanschris requesting a review on this one since I'm not sure its right.

If you decline the cookies:
* shows banner saying 'you have declined cookies' 
* button to click hide it (no auto disappear - hide must be clicked)

If you accept the cookies:
* reloads the page
* looks at the cookie that was set (which I've had to modify) checks the date it was created works out if its within **1 minute** of the current time - if it is it displays the banner.
* button to click hide it (no auto disappear - hide must be clicked)

I'm not sure if the accept is in-keeping with the rules of cookie setting since we will then in theory always be accessing that cookie on page load.

what do you think?

### DIFF
--- a/src/library/structure/CookieBanner/CookieBanner.styles.js
+++ b/src/library/structure/CookieBanner/CookieBanner.styles.js
@@ -10,7 +10,27 @@ export const Container = styled.div`
   box-sizing: border-box;
 `
 
+export const CookieHide = styled.button`
+    border: 0;
+    background: transparent;
+
+  ${props => props.theme.fontStyles}
+    ${props => props.theme.linkStyles}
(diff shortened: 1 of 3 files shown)


apricot13:
ok! I think this is done now - It checks to see if we set the cookiesAcceptedConformationBanner is dismissed and if it isn’t updates it when it is. bit better than looking at dates!

apricot13:
Tuesday, 30 March 2021 3:15 pm - decision made to delay until after vesting day!

apricot13:
post vesting day I have rebased this ready to be merged! happy for me to merge @evanschris @deesonlee ?

PR #24, opened 2021-03-25
NOR-135 Iterate news listing component design to match new design

apricot13:


apricot13:
@evanschris 
Its harder to make them not as square - on the design each card is 800px wide but in the code theyre ~700px. When I increase the width of the image because of excerpt it just doesn't quite work as well?


Desktop:
image

iPad
image

Phone
image

PR #43, opened 2021-07-12
removing skip to main content from Header and HomeHero

AnthonyTaylor:
To remove duplicate skip to main content link on homepage and remove unused import


apricot13 (commented):
Question - do we want to remove this? We have the ability to add extra ones in the admin but if this option is accidentally deleted from the admin it would be really bad? Does the admin section have a 'you cant delete this' option to prevent it? and to ensure the skip to content option always comes first?

AnthonyTaylor:
The component SkipToMainContent is pulled in directly into the frontend now instead of being in the Header and HomeHero  pieces, I had to take it out of the Header component as my new accessibility links were in the wrong place.

apricot13 (approved):

PR #44, opened 2021-07-13
#42 pagination fix

barka14:
Added in some code that used map to go through params to look for page param, and set to remove the page number from the search bar upon any changes to the search term. Also added in functionality to remove page numbers when changing search filters in news items or searching for different news items. 

To test in story book
comment out line 19 in url-helpers.js then open up examples--search-results change to another page and then change the default search term to something else and the page number should be deleted in the url bar. once happy then open up example-news-article-list and again move to a new page, then check the following remove the page numbers from the url, changing the sort by list, unticking an article type and changing the dropdown for the services


AnthonyTaylor (approved):
Approved. this will be deployed to NPM and be on the staging site by the end of the week as I will be doing another frontend deployment.

PR #71, opened 2021-11-01
Autocomplete component

DamienBurkeWN:
Relates to https://github.com/FutureNorthants/northants-design-system/issues/69

### Testing
1. Within Storybook you'll see there is a new Autocomplete input under **Components**, have a play with the stories and verify there are no visual problems and that suggestions display in accordance with what you type.
2. The component has been plugged in to the existing Searchbar structure so play around with that too; it would be a good idea to have the live copy of Storybook up to compare for visual differences. You will find some small differences in the focus outlining, these have been discussed with Lawrence and he's happy they match the gov.uk sites (black inner border stays) rather than the existing design system behaviour (black inner border disappears).
3. The Searchbar is used within the Header structure so try playing with that too.
4. The Searchbar is also used within the HomeHero structure so try that out as well. The stories for this one have an empty array for the search suggestions so you'll need to enter some of your own.


util/create-component.js
```diff
@@ -11,7 +11,7 @@ if (!componentName) {
 
 console.log("Creating Component Templates with name: " + componentName);
 
-const componentDirectory = `./src/${componentName}`;
```
DamienBurkeWN, line 14: Changed these as I found the readme instructions for creating a new component ended up with the created files in the wrong place.

src/library/components/Autocomplete/Autocomplete.types.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * See the stories file for explanation of each property or view the control in Storybook
+ */
+export interface AutocompleteProps {
+    name: string;
```
apricot13, line 5: in storybook the block comment above these becomes the description for the description field in the docs section:

image

src/library/components/Autocomplete/Autocomplete.styles.js
```diff
@@ -0,0 +1,48 @@
...
+    width: 21rem;
+    margin-left: 0rem;
+    margin-top: -1.3rem;
+    font-family: ${props => props.theme.theme_vars.fontstack};
+`;
+
+/**
+ * Suggestion items that display within the suggestion list
+ */
+export const AutocompleteSuggestionItem = styled.li`
```
apricot13, line 24: for the styling we try to follow the gov uk styles as much as possible. 
For this I would use the following. 
image

src/library/components/Autocomplete/Autocomplete.styles.js
```diff
@@ -0,0 +1,48 @@
...
+    padding: 0.5rem;
+    cursor: pointer;
+    border: 1px solid darkgray;
```
apricot13, line 27: all colours should be sourced from the props, so the border here will be the same as the colour used by the input styles. Same with any radius, font size or spacingSizes we set - each theme has it slightly different.

src/library/components/Autocomplete/Autocomplete.styles.js
```diff
@@ -0,0 +1,48 @@
...
+    font-size: 1rem;
+    text-align: left;
+    border-radius: 5px; !important
```
apricot13, line 30: you should never need to use `!important` - you also have a syntax error here though it will go away with the govuk restyling anyway!

apricot13 (changes requested):

util/create-component.js
apricot13: yeh it puts them outside the library by default so you dont have to go digging too far!

src/library/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,142 @@
+import React, {useState} from "react";
+import { AutocompleteProps } from "./Autocomplete.types";
+import { ErrorText} from "../Input/Input.styles";
+import * as Styles from "./Autocomplete.styles";
+import Downshift from "downshift";
+import { action } from '@storybook/addon-actions';
```
apricot13, line 6: this very likely will cause all sorts of issues when this is exported out and used in the frontend - you can test it locally in the frontend using yalc (instructions are in the readme). instead of using the frontend I sometimes use this as a test environment for single components: https://github.com/apricot13/FN-IE11-react-test-env

apricot13 (changes requested):

src/library/components/Autocomplete/Autocomplete.types.ts
DamienBurkeWN: Didn't work for me, the description fields are being picked up from the ArgTypes export in the stories file.

src/library/components/Autocomplete/Autocomplete.tsx
DamienBurkeWN: Good spot, should have thought that one through - action handling now moved to the stories file where it should have been.

src/library/components/Autocomplete/Autocomplete.styles.js
DamienBurkeWN: Try now?

apricot13: image
ooh much better! Theres an issue with the borders though they're all different widths, the auto complete box is 1px wider on the right hand side too it looks like.

src/library/components/Autocomplete/Autocomplete.types.ts
apricot13: ah ok - would you mind addin gthem to the types.ts too - since every other component is in there. 
we use the .types.ts file in the npm package in the frontend too - It may help with code completion in there too (but I dont actually know that for sure!)

DamienBurkeWN: Done - it appears Storybook doesn't update the displayed docs unless stopped and restarted!

src/library/components/Autocomplete/Autocomplete.styles.js
DamienBurkeWN: Fixed the width, don't see a difference in borders?

src/index.ts
```diff
@@ -23,8 +24,14 @@ export { Accordion, BlockQuote, CallToAction, Divider, DownloadableFiles, Warnin
 
 
 // Structure
-export { AccessibleLinks, AlertBanner, AlertBannerService, Breadcrumbs, CookieBanner, Footer, Header, HomeHero, MemorialHero, HomeUnitarySection, DisplayDate, NewsArticleDate, NewsArticleImage, NewsArticleFilterAccordion, NewsArticleList, NewsArticleListHeader, NewsArticleOldBanner, NewsArticleFeaturedBlock, SkipToMainContent, PhaseBanner, PageMain, PageTitle, PageSidebar, PageWithSidebarContainer, PageWrapper, PromoBanner, Searchbar, SearchResultsList, SectionLinks, SectionLinksSidebar, SectionLinksMobileContents, ServicesLinksList, SignpostLinks, MaxWidthContainer} from "./library/structure/PageStructures";
+export {
```
DamienBurkeWN, line 27: Just reformatted to get rid of the enormous long line

src/library/structure/NewsArticleFeaturedBlock/NewsArticleFeaturedBlock.tsx
```diff
@@ -5,7 +5,7 @@ import LazyImage from "react-lazy-progressive-image";
 import { NewsArticleFeaturedBlockProps } from "./NewsArticleFeaturedBlock.types";
 import * as Styles from "./NewsArticleFeaturedBlock.styles";
 import NewsArticleDate from "../NewsArticleDate/NewsArticleDate";
-import { Heading } from "../../..";
+import Heading from "../../components/Heading/Heading";
```
DamienBurkeWN, line 8: Referencing the component directly rather than via the index.ts, which creates a circular reference

src/library/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,159 @@
...
+            {isErrored && errorText ? <ErrorText>{errorText}</ErrorText> : ""}
+            <div
+              {...getRootProps(undefined, { suppressRefError: true })}
+            >
+              <Styles.AutocompleteTextInput {...getInputProps({ 
+                name: name,
+                value: inputValue,
+                placeholder: placeholder, 
+                isErrored: isErrored,
+                isOpen: isOpen && filteredsuggestions.length > 0,
```
DamienBurkeWN, line 117: A tweak so the squared off lower border doesn't become near-permanent once the list has displayed once

AnthonyTaylor (approved):

PR #72, opened 2021-11-03
Styles for autocomplete

apricot13:
Small PR for the auto complete - there was an issue with the margin-top making it look like a 3px border and the thicker bottom border was from the list items inside bottom border. also I brought it up a bit so needed to do some radius swapping though thinking now you could just bring it up a few px more and that wouldn't be an issue

### DIFF
--- a/src/library/components/Autocomplete/Autocomplete.styles.js
+++ b/src/library/components/Autocomplete/Autocomplete.styles.js
@@ -9,7 +9,6 @@ export const AutocompleteInput = styled(StyledInput)`
     margin-bottom: 0px;
     border-bottom-left-radius: ${props => props.isOpen ? "0px" : "2px"};
     border-bottom-right-radius: ${props => props.isOpen ? "0px" : "2px"};
-    /* background: ${props => props.isOpen ? "red" : "2px"} */
 `;
 
 /**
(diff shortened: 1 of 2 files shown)


src/library/components/Autocomplete/Autocomplete.styles.js
```diff
@@ -6,6 +6,10 @@ import { StyledInput } from "../Input/Input.styles";
  */
 export const AutocompleteInput = styled(StyledInput)`
     width: 20rem;
+    margin-bottom: 0px;
+    border-bottom-left-radius: ${props => props.isOpen ? "0px" : "2px"};
+    border-bottom-right-radius: ${props => props.isOpen ? "0px" : "2px"};
+    /* background: ${props => props.isOpen ? "red" : "2px"} */
```
DamienBurkeWN, line 12: Guessing you didn't mean to leave this bit in?

src/library/components/Autocomplete/Autocomplete.tsx
```diff
@@ -105,7 +105,8 @@ const Autocomplete: React.FunctionComponent<AutocompleteProps> = ({
               <Styles.AutocompleteInput {...getInputProps({ 
                 name: name,
                 placeholder: placeholder, 
-                isErrored: isErrored
+                isErrored: isErrored,
+                isOpen: isOpen
```
DamienBurkeWN, line 109: ...and that isOpen isn't needed here as a result too.

DamienBurkeWN (changes requested):

src/library/components/Autocomplete/Autocomplete.tsx
apricot13: isOpen is still needed for the borders 👍

DamienBurkeWN (approved):

PR #74, opened 2021-11-09
fixes insufficient colour contrast in search box placeholder

AnthonyTaylor:


DamienBurkeWN (approved):
Works, but is it worth moving those hard-coded colour values into the theme?

PR #85, opened 2021-12-16
Feature/video component

chrisrhymes:
Video component that allows a YouTube video to be embedded with the YouTube id. The container is set to 16 by 9 ratio, so it is responsive to the device.


src/library/components/Video/Video.styles.js
```diff
@@ -0,0 +1,17 @@
+import styled from 'styled-components';
+
+export const VideoContainer = styled.figure`
+  display: block;
+  padding-top: 56.25%;
```
DamienBurkeWN, line 5: What's the magic that makes this figure work? I thought 50 would have been right but it clearly isn't.

src/library/components/Video/Video.styles.js
```diff
@@ -0,0 +1,17 @@
...
+
+  iframe {
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+  }
```
DamienBurkeWN, line 16: Shouldn't we be turning off any border here?

src/library/components/Video/Video.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
```
DamienBurkeWN, line 2: Tidy - no need for this blank line

src/library/components/Video/Video.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { VideoProps } from './Video.types';
+import * as Styles from './Video.styles';
+
+const Video: React.FC<VideoProps> = ({ youtube_id }) => (
```
DamienBurkeWN, line 6: Could do with a comment block here describing the component (shows up in Storybook), and I've been using .FunctionComponent rather than .FC for clarity/readability, I think it helps when new to React that you know explicitly what it's doing

src/library/components/Video/Video.tsx
```diff
@@ -0,0 +1,18 @@
...
+  <Styles.VideoContainer data-testid="Video">
+    <iframe
+      src={`https://www.youtube.com/embed/${youtube_id}?rel=0`}
+      title="YouTube video player"
+      frameborder="0"
```
DamienBurkeWN, line 11: This isn't a valid attribute any more is it? Should be handled in the css?

src/library/components/Video/Video.tsx
```diff
@@ -0,0 +1,18 @@
...
+      allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
+      allowfullscreen
```
DamienBurkeWN, line 13: Another invalid attribute - should be 'fullscreen' in the allow list above

DamienBurkeWN (changes requested):
Nice and simple but 3 more things over and above the comments. 

First off I presume including package-lock.json was a mistake, there's no package changes.

Second, I assume there's a slice to follow this which will either make use of this component or a link to the video if the user said no to cookies?

Third, could this component be more generic - e.g. support Vimeo or other video platforms, rather than being specifically Youtube? I'm thinking whilst the current two councils have youtube channels, microsite owners may want to use other or multiple platforms.

src/library/components/Video/Video.styles.js
DamienBurkeWN: Ah never mind, 16/9 - gotcha!

src/library/components/Video/Video.tsx
chrisrhymes: Didn't know about the comment, thanks. I've added the comment and used FunctionComponent.

src/library/components/Video/Video.tsx
chrisrhymes: I've applied the border style to the styles.

src/library/components/Video/Video.tsx
chrisrhymes: I have added 'fullscreen' to the allow.

src/library/components/Video/Video.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { VideoProps, VideoProvider } from './Video.types';
+import * as Styles from './Video.styles';
+
+/**
+ * A responsive video embed for YouTube or Vimeo
+ * @param param0
```
DamienBurkeWN, line 7: Just a whoops on the needless param and return in the docblock

DamienBurkeWN (approved):
Spot on. Just a couple of comment lines to kill and all done I think.

DamienBurkeWN (approved):

chrisrhymes:
Refactored the video component to display a link when the cookies are not accepted and added tests.

src/library/components/Video/Video.stories.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
```
DamienBurkeWN, line 1: As it's to be directly used as a slice rather than being a component in a slice, I think all this lot needs to be moved into src/library/slices

src/library/components/Video/Video.tsx
```diff
@@ -0,0 +1,73 @@
+import React, { useEffect, useState } from 'react';
+import { VideoProps, VideoProvider } from './Video.types';
+import * as Styles from './Video.styles';
+import { cookieName, getCookie } from '../../structure/CookieBanner/CookieHelpers';
+import Button from '../Button/Button';
```
DamienBurkeWN, line 5: This import isn't used

DamienBurkeWN (changes requested):
Looks good, as per comment needs moving from components to slices. Also still needs the package-lock.json file taking out.

DamienBurkeWN (changes requested):
image

DamienBurkeWN (approved):
Nice

PR #86, opened 2021-12-16
Bugfix/jest

chrisrhymes:
- Fix the existing Jest test suite
- Add GitHub action to run tests on push


DamienBurkeWN (approved):
Excellent :clap:

.storybook/main.js
```diff
@@ -1,23 +1,21 @@
...
+    "@react-theming/storybook-addon", // https://github.com/react-theming/storybook-addon
     "@storybook/addon-links",
     "@storybook/addon-essentials",
-    "@storybook/addon-a11y",              // https://github.com/storybookjs/storybook/tree/master/addons/a11y
-    "@etchteam/storybook-addon-status",    // https://storybook.js.org/addons/@etchteam/storybook-addon-status/
-    "@storybook/addon-ie11"
+    "@storybook/addon-a11y", // https://github.com/storybookjs/storybook/tree/master/addons/a11y
+    "@etchteam/storybook-addon-status", // https://storybook.js.org/addons/@etchteam/storybook-addon-status/
+    "@storybook/addon-ie11",
+    "@storybook/addon-docs",
```
chrisrhymes, line 12: Needed to add the @storybook/addon-docs after updating other packages. This also required typescript ^4, rather than ^3 set previously in the package.json.

package.json
```diff
@@ -75,9 +77,9 @@
     "eslint-plugin-react-hooks": "^4.2.0",
     "eslint-plugin-storybook": "^0.1.1",
     "identity-obj-proxy": "^3.0.0",
-    "jest": "^24.9.0",
+    "jest": "^26",
     "jsuri": "^1.3.1",
-    "node-sass": "^4.14.1",
+    "node-sass": "^6",
```
chrisrhymes, line 82: Had to update to v6 to work with newer version of node, along with sass-loader.

src/library/components/PromotedLinks/PromotedLinks.test.tsx
```diff
@@ -1,27 +1,42 @@
...
+        },
+      ],
+      oneCol: true,
     };
   });
 
-  const renderComponent = () => render(<PromotedLinks {...props} />);
+  const renderComponent = () =>
+    render(
+      <ThemeProvider theme={west_theme}>
```
chrisrhymes, line 30: Components need to be wrapped in the ThemeProvider so they have access to the theme config.

chrisrhymes:
I have reverted the commits that used the newer version of chrome and updated the GitHub action to now use node 14. It should now also cache the npm install, hopefully speeding up future runs too.

PR #89, opened 2021-12-22
Bugfix/empty search term

chrisrhymes:
- Prevents user searching without a search term (https://github.com/FutureNorthants/northants-website/issues/192).
- Add tests for Searchbar, including displaying suggestions and displaying error state without a search term

### DIFF
--- a/src/library/structure/Searchbar/Searchbar.test.tsx
+++ b/src/library/structure/Searchbar/Searchbar.test.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import {
   render,
   fireEvent,
+  getByPlaceholderText,
   getByRole,
   getByTestId,
   waitFor,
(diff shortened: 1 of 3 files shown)


src/library/structure/Searchbar/Searchbar.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const listbox: HTMLElement = getByRole(component, "listbox");
+
+    expect(listbox.children.length).toBe(1);
+  });
+
+  it("should display the error state when no search term provided", () => {
+    fireEvent.click(searchButton);
+
+    // #9D0B1D is west_theme colours.negative
+    expect(searchInput).toHaveStyle("border: solid #9D0B1D");
```
DamienBurkeWN, line 64: Is it possible to pick it up by name from the theme rather than hardcoded? A theme alteration will break the test at the moment.

src/library/structure/Searchbar/Searchbar.tsx
```diff
@@ -39,23 +40,27 @@ const Searchbar: React.FunctionComponent<SearchbarProps> = ({
   }
 
   /**
-   * Run after every render if the dependencies in the array change; we use this to check if submit 
+   * Run after every render if the dependencies in the array change; we use this to check if submit
    * has been set to true and if so, we submit the search.
    */
   useEffect(
     () => {
+      setIsErrored(false);
```
DamienBurkeWN, line 48: This triggers a re-render (2 renders per keystroke vs 1) regardless of the previous value of _isErrored_, would suggest adding `if (isErrored)` before it.

src/library/structure/Searchbar/Searchbar.tsx
```diff
@@ -65,29 +70,52 @@ const Searchbar: React.FunctionComponent<SearchbarProps> = ({
    */
   const handleSubmit = (event) => {
     if (event) event.preventDefault();
+
+    if (inputSearchTerm == "") {
+      setIsErrored(true);
```
DamienBurkeWN, line 75: Again this triggers a re-render regardless of previous value of isErrored so doing it conditionally is better.

DamienBurkeWN (changes requested):

DamienBurkeWN (approved):
AAA+ would review again

PR #95, opened 2022-01-12
Center align the AlertBannerService content

chrisrhymes:
Resolves this issue: https://github.com/FutureNorthants/northants-website/issues/213

The content has been kept to a maximum of 750px, but centred in the alert instead of


WestNorthantsRob:
Chris, I'm not sure having it centred works from an accessibility point of view, If someone has enlarged their screen they may not realise there is something in the box.

chrisrhymes:
Thanks for the feedback @WestNorthantsRob. I have updated it to be fullwidth. 

Please see the updated [Service Page](https://deploy-preview-95--northants-design-system.netlify.app/?path=/story/page-examples-service-page-example--example-page-with-service-alert) and [Signpost Page](https://deploy-preview-95--northants-design-system.netlify.app/?path=/story/page-examples-signpost-page-examples--single-example-with-service-alert) examples

DamienBurkeWN (approved):
👍

PR #97, opened 2022-01-12
Set searchbar label value and add prop to hide it

DamienBurkeWN:
### Testing
View stories: home page, header, home hero, searchbar. Inspect the html and verify the nearby label is not empty, and is visually hidden.

### DIFF
--- a/src/library/components/Autocomplete/Autocomplete.stories.tsx
+++ b/src/library/components/Autocomplete/Autocomplete.stories.tsx
@@ -24,6 +24,9 @@ export default {
       labelText: {
         table: { category: "Input control" }
       },
+      labelHidden: {
+        table: { category: "Input control" }
+      },
       value: {
         table: { category: "Input control" }
       },
@@ -99,22 +102,25 @@ const Template: Story<AutocompleteProps> = (args) => <SBPadding><div style={{wid
 /**
  * Empty, i.e. default state of the Autocomplete component
  */
-export const Empty = Template.bind({});    
-Empty.args = {
+export const EmptyWithHiddenLabel = Template.bind({});    
+EmptyWithHiddenLabel.args = {
   ...CommonArgs,
   id: 'empty',
-  name: 'Empty'
+  name: 'Empty',
+  labelHidden: true,
 }
 
 /**
  * Autocomplete component with some text entered that doesn't match any suggestion
  */
-export const NoMatchingSuggestionEnlarged = Template.bind({});    
-NoMatchingSuggestionEnlarged.args = {
+export const NoMatchingSuggestionEnlargedWithHiddenLabel = Template.bind({});    
+NoMatchingSuggestionEnlargedWithHiddenLabel.args = {
   ...CommonArgs,
   id: 'changed',
   name: 'Changed',
-  value: 'Banana'
+  value: 'Banana',
+  isLarge: true,
+  labelHidden : true,
 };
 
 /**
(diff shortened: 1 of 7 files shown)


src/library/components/Autocomplete/Autocomplete.styles.js
```diff
@@ -1,79 +1,116 @@
-import styled from "styled-components";
```
DamienBurkeWN, line 1: Mostly prettier...

src/library/components/Autocomplete/Autocomplete.styles.js
```diff
@@ -1,79 +1,116 @@
...
+const hideLabel = props => {
+  if (props.labelHidden === true) {
+    return VisuallyHidden;
+  }
+}
+
 export const AutocompleteLabel = styled.label`
-    font-family: ${props => props.theme.theme_vars.fontstack};
+  font-family: ${props => props.theme.theme_vars.fontstack};
+  ${props => hideLabel}
```
DamienBurkeWN, line 38: This is the real change, see hideLabel above

src/library/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -24,6 +24,9 @@ export default {
       labelText: {
         table: { category: "Input control" }
       },
+      labelHidden: {
```
DamienBurkeWN, line 27: New control so the associated label can be hidden

src/library/components/Autocomplete/Autocomplete.tsx
```diff
@@ -112,7 +113,7 @@ const Autocomplete: React.FunctionComponent<AutocompleteProps> = ({
           getRootProps
         }) => (
           <div>
-            <Styles.AutocompleteLabel {...getLabelProps()}>{labelText}</Styles.AutocompleteLabel>
+            <Styles.AutocompleteLabel {...getLabelProps()} labelHidden={labelHidden}>{labelText ? labelText : placeholder }</Styles.AutocompleteLabel>
```
DamienBurkeWN, line 116: Now uses placeholder text if no labeltext specified

chrisrhymes:
Would something like this be useful for testing potential accessibility issues? https://github.com/nickcolley/jest-axe

src/library/components/Autocomplete/Autocomplete.tsx
```diff
@@ -13,6 +13,7 @@ const Autocomplete: React.FunctionComponent<AutocompleteProps> = ({
   id = "autocomplete",
   name = "autocomplete",
   labelText,
+  labelHidden = false,
```
chrisrhymes, line 16: Is autocomplete used anywhere else right now?

DamienBurkeWN: Not yet

src/library/components/Autocomplete/Autocomplete.tsx
chrisrhymes: The label could do with some styling when it is visible, such as some padding between it and the input.

chrisrhymes:
Good opportunity to check that the label is set, but visible/ not visible with a test on the autocomplete component.

src/library/components/Autocomplete/Autocomplete.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+} from './Autocomplete.storydata';
+import { west_theme } from '../../../themes/theme_generator';
+import { ThemeProvider } from 'styled-components';
+
+describe('Autocomplete component', () => {
+  it('should render a hidden label and empty input', () => {
+    const renderComponent = () =>
+      render(
+        <ThemeProvider theme={west_theme}>
+          <Autocomplete {...EmptyWithHiddenLabelData} />
```
chrisrhymes, line 18: Great idea to use the story data in the tests!

chrisrhymes (approved):

PR #100, opened 2022-01-18
Prevent max height when header is hidden

chrisrhymes:
There is a max height set on mobile and a view more button appears, but the view more button only appears if `hideHeader == true`. Top Services hides the header so the view more button doesn't appear and the max-height is applied, cutting the content.

It now passes hideHeader to the Container styles. If it's true it will not set a max height. 

If the header isn't hidden, then the view more button will appear and the max-height will be applied on mobile.


DamienBurkeWN (approved):
👍

PR #101, opened 2022-01-18
Make AlertBannerService title optional

chrisrhymes:
If the title is not set, then it will not include the heading.

- Added story for alert without a title
- Added test to ensure heading is not included when title is empty

To test, run `npm run test AlertBannerService`

### DIFF
--- a/src/library/structure/AlertBannerService/AlertBannerService.tsx
+++ b/src/library/structure/AlertBannerService/AlertBannerService.tsx
@@ -13,7 +13,7 @@ const AlertBannerService: React.FunctionComponent<AlertBannerServiceProps> = ({
 }) => (
   <Styles.Container alertType={alertType} data-testid="AlertBannerService">
     <Styles.InnerContainer data-testid="AlertBannerServiceInner">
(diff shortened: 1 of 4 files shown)


src/library/structure/AlertBannerService/AlertBannerService.tsx
```diff
@@ -13,7 +13,7 @@ const AlertBannerService: React.FunctionComponent<AlertBannerServiceProps> = ({
 }) => (
   <Styles.Container alertType={alertType} data-testid="AlertBannerService">
     <Styles.InnerContainer data-testid="AlertBannerServiceInner">
-      <Heading text={title} />
+      {title && <Heading text={title} />}
```
DamienBurkeWN, line 16: This works for empty string or no title at all but I've seen them use just spaces in a title, and that would provoke an accessibility warning for empty header again. Replace with `title?.trim() &&` perhaps?

DamienBurkeWN (changes requested):
One suggestion

DamienBurkeWN (approved):

PR #102, opened 2022-01-20
Feature/update storybook 6.4

chrisrhymes:
- Update to storybook 6.3.* from 6.1
- Update to use webpack5 and update config
- Update to React 17, node-sass 7, typescript 4.


DamienBurkeWN (changes requested):
Functionally this is working now but there's a regression on the design system version number - it's 0.1.135 in package.json despite master being at 0.1.136, and github doesn't seem to have spotted this. When I run npm install on this branch I end up with a changed package-lock.json too.

DamienBurkeWN (approved):

PR #103, opened 2022-01-21
Replace react-theming and update to SB 6.4

chrisrhymes:
This allows for update to Storybook 6.4. 

Remove @react-theming/storybook-addon package and replace with global theme variable using [this example](https://storybook.js.org/docs/react/essentials/toolbars-and-globals#create-a-decorator) in the storybook docs. 

You can now select the theme from the top menu bar. The main difference is the theme variables are no longer visible in the bottom bar.

The addDecorator helper has also been removed as it was [deprecated in storybook 6.0](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-addparameters-and-adddecorator).


DamienBurkeWN (approved):
Niiiiice

PR #108, opened 2022-02-01
Full width container & accordion changes

chrisrhymes:
A couple of small changes from discussion with Lawrence last week. 

- Reduce the bottom padding of the full width container. Content inside usually has it's own bottom margin.
- Remove the 'Read more' links in the accordions by default
- Updated Accordion story to show with 'Read more' as it's now hidden by default


DamienBurkeWN (approved):
👍

PR #111, opened 2022-02-03
Reduce H2 margin

DamienBurkeWN:
### Testing
1. Look at Heading level 2. 25 pix top margin? Good.


chrisrhymes (approved):
Looks ok on storybook. I am unsure why h2 styles are defined in the theme and then declared with different values in the heading component, but that's a question for another day...

PR #112, opened 2022-02-03
Feature/image component

chrisrhymes:
- Add image slice
- Update storybook
- Update service page example with an image

The addition of a responsive image container, using the lazy image, with optional caption. There are two available image ratio's, 4by3 and 16by9. The design is in Trello and the caption has different background colours based on the selected theme.


DamienBurkeWN (approved):
👍 but needs merge conflict sorted

PR #113, opened 2022-02-03
Add Promotions slice

DamienBurkeWN:
### Testing
Have a look at the content page demo, down the bottom, for example use of the slice and also the stories of the slices/Promotions component.

As it's an assembly of two existing structure components I've not written a test for this one - need to revisit the existing components to give them tests first I think.


src/library/structure/PromoBlock/PromoBlock.styles.js
```diff
@@ -1,5 +1,5 @@
-import styled from "styled-components";
```
DamienBurkeWN, line 1: Mostly prettier but also specifying theme font styling in the PromoContent div rather than relying on the container

chrisrhymes:
Do we want users to be able to add multiple promo blocks or just add one promo banner on a page? I suppose that's more up to the CMS setup than this PR.

DamienBurkeWN:
I suspect it will be one block on a page but I can't actually see how we can enforce that in the CMS yet without also enforcing its position on the page, which I suspect won't be acceptable.

chrisrhymes (approved):
Good idea reusing the existing components. Should make it easier to maintain in future.

PR #121, opened 2022-02-16
Feature/roads highways icons

chrisrhymes:
Add new Roads and Highways icons. I also took this opportunity to refactor the Section Links and extract the icon and heading into its own component. It is now a single link, instead of two separate links. This caused issues with screen readers reading out the link twice.


src/library/structure/ServicesLinksList/ServicesLinksList.test.tsx
```diff
@@ -56,8 +56,8 @@ describe('Services Links List', () => {
     expect(header).toHaveTextContent('Order services by');
 
     const links = queryAllByRole('link');
-    // Quick links display: none on mobile and each service has icon and text link
-    expect(links.length).toBe(4);
+    // Quick links display: none on mobile so they won't be included
+    expect(links.length).toBe(2);
```
chrisrhymes, line 61: Previously, the icon and title were two separate links. Now they are combined, hence the change from 4 to 2.

DamienBurkeWN (approved):
♿ 👍

PR #130, opened 2022-02-18
Bugfix/google map styles overwritten

chrisrhymes:
Changed to styled.figure and it seems to resolve the issue with the map going fullscreen when there is a WarningTextDisclaimer on the page. 

Added Google map to Content Page example to try and debug, but the bug doesn't occur in the design system, only the frontend.

Also updated dependencies (since yesterday's update...)

## To Test
- `yalc publish`
- cd to frontend and `yalc add northants-design-system && rm -r .next/cache && yarn install && yard dev`
- Visit page with Google Map slice and a WarningTextDisclaimer

### DIFF
--- a/src/library/pages/ContentPage/ContentPage.storydata.ts
+++ b/src/library/pages/ContentPage/ContentPage.storydata.ts
@@ -1,5 +1,7 @@
 /** Example story data for Content Page example */
 
+import { GoogleMapProps } from '../../slices/GoogleMap/GoogleMap.types';
+
 export const smallTable = {
   caption: 'Month you apply',
   headings: ['Date', 'Rate for vehicles', 'Rate for bicycles'],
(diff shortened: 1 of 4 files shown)


src/library/pages/ContentPage/ContentPage.storydata.ts
```diff
@@ -14,3 +16,12 @@ export const largeTable = {
   caption: 'Example table with lots of data',
   headings: ['', 'Band A', 'Band B', 'Band C', 'Band D', 'Band E', 'Band F', 'Band G', 'Band H'],
 };
+
+export const googleMapData: GoogleMapProps = {
```
DamienBurkeWN, line 20: It would be better to import from the slice's storydata file rather than add more data here.

DamienBurkeWN (changes requested):

DamienBurkeWN (approved):
Ahhh less diff 👍

PR #134, opened 2022-02-22
Fix Google Map embed regex check for Safari

DamienBurkeWN:
### Testing
- View google map slice in Safari. Works? Good.


chrisrhymes (approved):
👍

PR #137, opened 2022-02-23
Make PromoBanner image clickable, add test

DamienBurkeWN:
### Testing
- Check out library/structure/promotional banner - verify image is now clickable and goes to same URL as call to action link
- `yarn test PromoBanner`

🪄 Reserving right to add protective div if it blows up on the frontend.

### DIFF
--- a/src/library/structure/PromoBanner/PromoBanner.stories.tsx
+++ b/src/library/structure/PromoBanner/PromoBanner.stories.tsx
@@ -1,19 +1,44 @@
-import React from "react";
+import React from 'react';
 import { Story } from '@storybook/react/types-6-0';
-import PromoBanner from "./PromoBanner";
-import { PromoBannerProps } from "./PromoBanner.types";
+import PromoBanner from './PromoBanner';
+import { PromoBannerProps } from './PromoBanner.types';
 import { SBPadding } from '../../../../.storybook/SBPadding';
-import { PromoBannerContent, PromoBannerData } from "./PromoBanner.storydata";
+import { PromoBannerContent, PromoBannerData } from './PromoBanner.storydata';
 
 export default {
-    title: "library/Structure/Promo Banner",
-    parameters: {
-      status: {
-        type: 'stable', // 'beta' | 'stable' | 'deprecated' | 'releaseCandidate'
-      }
+  title: 'library/Structure/Promo Banner',
+  component: PromoBanner,
+  parameters: {
+    status: {
+      type: 'stable', // 'beta' | 'stable' | 'deprecated' | 'releaseCandidate'
     },
+  },
+  argTypes: {
+    image1440x810: {
+      table: { category: 'Image' },
+    },
+    image144x81: {
+      table: { category: 'Image' },
+    },
+    title: {
+      table: { category: 'Content' },
+    },
+    children: {
+      table: { category: 'Content' },
+    },
+    ctaText: {
+      table: { category: 'Call to Action' },
+    },
+    ctaUrl: {
+      table: { category: 'Call to Action' },
+    },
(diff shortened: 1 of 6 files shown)


src/library/structure/PromoBanner/PromoBanner.stories.tsx
```diff
@@ -1,19 +1,44 @@
...
-        type: 'stable', // 'beta' | 'stable' | 'deprecated' | 'releaseCandidate'
-      }
+  title: 'library/Structure/Promo Banner',
+  component: PromoBanner,
+  parameters: {
+    status: {
+      type: 'stable', // 'beta' | 'stable' | 'deprecated' | 'releaseCandidate'
     },
+  },
+  argTypes: {
```
DamienBurkeWN, line 16: Mostly prettier but also adding the categories for controls

src/library/structure/PromoBanner/PromoBanner.storydata.tsx
```diff
@@ -3,13 +3,14 @@
  */
 import React from "react";
 
-export const PromoBannerContent = <p>We’re looking for people to come forward and help deliver the Covid-19 vaccination programme in North Northamptonshire.</p>;
+export const PromoBannerContentText = "We’re looking for people to do this amazing thing in an amazing manner, during an amazing time of amazingness.";
```
DamienBurkeWN, line 6: Moved into a string so it can be picked up and referenced in the test

src/library/structure/PromoBanner/PromoBanner.styles.js
```diff
@@ -32,17 +32,16 @@ export const Container = styled.div`
...
-    background-image: url("${(props) => props.img}");
-    background-size: cover;
-    background-position: center;
-    min-height: 200px;
-
-    @media screen and (min-width: ${(props) => props.theme.theme_vars.breakpoints.m}){
-        width: 50%;
-        min-height: 300px;
-    }
+export const Image = styled.a`
```
DamienBurkeWN, line 35: The actual change 😆

src/library/structure/PromoBanner/PromoBanner.tsx
```diff
@@ -1,36 +1,37 @@
-import React from "react";
```
DamienBurkeWN, line 1: Mostly prettier

src/library/structure/PromoBanner/PromoBanner.tsx
```diff
@@ -1,36 +1,37 @@
...
-    </Styles.Container>
+  <Styles.Container>
+    <LazyImage
+      src={image1440x810}
+      placeholder={image144x81}
+      visibilitySensorProps={{
+        partialVisibility: true,
+      }}
+    >
+      {(src) => <Styles.Image img={src} href={ctaUrl} title={ctaText} />}
```
DamienBurkeWN, line 27: Adding href and title now it's an anchor not a div

chrisrhymes: Just thinking might be worth changing Image to ImageLink just so in future it's clearer it's not just an image?

chrisrhymes: Do we also need alt text for this background image somehow?

chrisrhymes (changes requested):
Just a couple of comments about the Image. The rest looks great.

src/library/structure/PromoBanner/PromoBanner.tsx
DamienBurkeWN: It's historically not had one hence why I've added the title attribute

PR #144, opened 2022-03-01
Remove mobile breadcrumbs 

chrisrhymes:
- Remove the mobile breadcrumbs
- Adjust spacing so that it has even spacing when breadcrumbs wrap on small screens
- Replace px sizes with theme spacingSizes
- Update test as no longer has mobile back link
- Wrap stories in MaxWidthContainer so styles better represent theme styles

### DIFF
--- a/src/library/structure/Breadcrumbs/Breadcrumbs.styles.js
+++ b/src/library/structure/Breadcrumbs/Breadcrumbs.styles.js
@@ -15,9 +15,19 @@ export const List = styled.ol`
 `;
 
 export const Crumb = styled.li`
-  display: inline-block;
   margin-right: ${(props) => props.theme.theme_vars.spacingSizes.small};
   margin-bottom: ${(props) => props.theme.theme_vars.spacingSizes.small};
+  display: none;
+
+  &:first-of-type,
+  &:last-of-type {
+    display: inline-block;
+  }
+
+  @media screen and (min-width: ${(props) => props.theme.theme_vars.breakpoints.m}) {
+    display: inline-block;
+  }
+
   &:last-of-type svg {
     display: none;
   }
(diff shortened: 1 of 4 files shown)


DamienBurkeWN (approved):
Does what it says on the tin so approved, but the page content examples show examples of breadcrumbs that have more than 2 levels, which as far as I can see is never the case on the website. I was going to suggest, based on this example:

https://deploy-preview-144--northants-design-system.netlify.app/?path=/story/page-examples-service-page-example--example-page-with-service-alert

...that we should introduce a 'short breadcrumb title' that could be used to reduce the amount of screen space used for breadcrumbs on mobile. E.g. in that example instead of:

`Home > 
Bin collection, recycling and waste > 
Find your bin collection day`

We'd have shorter titles specifically for mobile, giving us as an example:

`Home > Recycling and waste > Bin day`

...but I'm not sure it's necessary if we only ever have 2 lines maximum? Worth getting Lawrence involved?

DamienBurkeWN:
It'd be a rare user that pays that much attention I suspect!

src/library/structure/Breadcrumbs/Breadcrumbs.styles.js
```diff
@@ -3,50 +3,46 @@ import styled from 'styled-components';
...
+
+export const Crumb = styled.li`
+  margin-right: ${(props) => props.theme.theme_vars.spacingSizes.small};
+  margin-bottom: ${(props) => props.theme.theme_vars.spacingSizes.small};
   display: none;
-  margin-bottom: 20px;
+
+  &:first-of-type,
+  &:last-of-type {
+    display: inline-block;
```
chrisrhymes, line 24: Only show first and last breadcrumb on small. Show all on medium and up.

DamienBurkeWN (approved):
👍

PR #146, opened 2022-03-02
Update PostCodeSearch placeholder

chrisrhymes:
Lots of changes for a placeholder text change... 

Managed to figure out how to mock axios (after many hours trawling stack overflow) so the tests shouldn't need access to the real api to run. This allows us to control the response. 

The tests wouldn't work with `e.target.postcode.value` in handleSubmit so I updated it to use `e.target.elements.postcode.value`. 

There was also a bug where it activated the loading state when the postcode was empty so it never displayed the error message.

### DIFF
--- a/src/library/components/FormWithLine/FormWithLine.tsx
+++ b/src/library/components/FormWithLine/FormWithLine.tsx
@@ -1,39 +1,44 @@
+import React from 'react';
 
-import React from "react";
-
-import { FormWithLineProps } from "./FormWithLine.types";
-import * as Styles from "./FormWithLine.styles";
+import { FormWithLineProps } from './FormWithLine.types';
+import * as Styles from './FormWithLine.styles';
 
 /**
  * Form element - a container with a line on the left
  */
 const FormWithLine: React.FC<FormWithLineProps> = ({
-  lineColour = "#C6C6C6",
-  hideLine = false, 
+  lineColour = '#C6C6C6',
+  hideLine = false,
   formRole,
   formMethod,
   formURL,
   isError = false,
   errorSummary,
   onSubmit,
-  children
-  }) => {
-    return (
-      <Styles.Container>
-        <Styles.Line lineColour={lineColour} hideLine={hideLine} isError={isError} />
-        <Styles.Form onSubmit={onSubmit} hideLine={hideLine} role={formRole} method={formMethod} url={formURL}>
-          {errorSummary && <Styles.ErrorSummary>{errorSummary}</Styles.ErrorSummary>}
-          {children}
-        </Styles.Form>
-      </Styles.Container>
-    );
-  };
-  
+  children,
+}) => {
+  return (
+    <Styles.Container>
+      <Styles.Line lineColour={lineColour} hideLine={hideLine} isError={isError} />
+      <Styles.Form
+        onSubmit={onSubmit}
+        hideLine={hideLine}
+        role={formRole}
+        method={formMethod}
+        url={formURL}
+        data-testid="FormWithLine"
+      >
+        {errorSummary && <Styles.ErrorSummary>{errorSummary}</Styles.ErrorSummary>}
+        {children}
+      </Styles.Form>
+    </Styles.Container>
+  );
+};
 
 export default FormWithLine;
 
-
-{/* <form  class="form form--area-search"  data-area-search="">
+{
+  /* <form  class="form form--area-search"  data-area-search="">
                         <div data-form-title class="area-search__title">Enter your postcode</div>
                         <div data-form-info class="area-search__info">For example HP20 1UY</div>
                         <div data-response-text class="area-search__response-text"></div>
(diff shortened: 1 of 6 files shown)


src/library/components/FormWithLine/FormWithLine.tsx
```diff
@@ -1,39 +1,44 @@
...
+  return (
+    <Styles.Container>
+      <Styles.Line lineColour={lineColour} hideLine={hideLine} isError={isError} />
+      <Styles.Form
+        onSubmit={onSubmit}
+        hideLine={hideLine}
+        role={formRole}
+        method={formMethod}
+        url={formURL}
+        data-testid="FormWithLine"
```
chrisrhymes, line 29: Added data-testid as the fireEvent.click wouldn't work. This allowed me to use fireEvent.submit() in the test.

src/library/components/PostCodeSearch/PostCodeSearch.tsx
```diff
@@ -1,87 +1,89 @@
...
-        checkPostcode(e.target.postcode.value);
-      }
+  const clearData = () => {
+    setResponseData(defaultArray);
+    setCurrentPostcode('');
+    setAddressArray([]);
+  };
+
+  const handleSubmit = (e) => {
+    const enteredPostcode = e.target.elements.postcode.value;
```
chrisrhymes, line 66: Updated to use `elements` and created variable to reuse.

src/library/components/PostCodeSearch/PostCodeSearch.tsx
```diff
@@ -90,136 +92,204 @@ const PostCodeSearch: React.FC<PostCodeSearchProps> = ({
...
-
-                <FormButton type="submit" aria-label="Submit" text="Find" />
+                  <Styles.Label htmlFor="postcode">
+                    Enter your postcode
+                    <HintText
+                      text={themeContext.cardinal_name === 'north' ? 'For example NN16 0AP' : 'For example NN1 1DE'}
+                    />
+                    <Input
+                      type="text"
+                      placeholder="Search postcode"
```
chrisrhymes, line 194: The actual change, updating the placeholder to be unique when used on a page with a search bar.

src/library/components/PostCodeSearch/PostCodeSearch.tsx
```diff
@@ -1,87 +1,89 @@
...
+  };
+
+  const handleSubmit = (e) => {
+    const enteredPostcode = e.target.elements.postcode.value;
+    e.preventDefault();
+    setCurrentPostcode(enteredPostcode);
+    if (enteredPostcode === '') {
+      handleError(true, 'You need to enter a postcode');
+    } else {
+      setIsLoading(true);
```
chrisrhymes, line 72: The `setIsLoading(true)` was in the wrong place so you never saw the error message.

DamienBurkeWN (changes requested):
As well as the individual comments to look at, there's a behaviour change introduced here. If you enter an unknown or outside-Northants postcode, your input used to be retained in the input so that you could - for instance - easily correct a typo. Now it is discarded entirely so you have to re-enter from scratch, and have reduced feedback on whether it was your fault.

src/library/components/Input/Input.tsx
```diff
@@ -1,27 +1,17 @@
...
-    errorText,
-    name
-  }) => {
-    return (
-      <>
+const Input: React.FC<InputProps> = ({ type = 'text', placeholder = '', isErrored = false, errorText, name }) => {
+  return (
+    <>
       {errorText && <Styles.ErrorText>{errorText}</Styles.ErrorText>}
-      <Styles.StyledInput type={type} placeholder={isErrored ? "Enter a postcode" : placeholder} name={name} isErrored={isErrored} />
```
chrisrhymes, line 20: Not sure why a generic component has a hardcoded string for an error state, so reverted to always use the provided placeholder. The rest is prettier

src/library/components/PostCodeSearch/PostCodeSearch.styles.js
```diff
@@ -1,115 +1,119 @@
...
 
 export const LoadingContainer = styled.div`
-    padding-left: 30px;
-`
\ No newline at end of file
+  padding-left: 30px;
+`;
+
+export const FormContainer = styled.div`
+  display: ${(props) => (props.isLoading ? `none` : 'block')};
```
chrisrhymes, line 118: Added css to hide the form container when isLoading. This way it keeps the state rather than removing and adding the form after loading.

src/library/components/PostCodeSearch/PostCodeSearch.tsx
```diff
@@ -90,136 +92,204 @@ const PostCodeSearch: React.FC<PostCodeSearchProps> = ({
...
+              lineColour={themeContext.theme_vars.colours.grey_dark}
+            >
+              {isLoading && (
+                <Styles.LoadingContainer>
+                  <LoadingSpinner />
+                  <p>Loading...</p>
+                </Styles.LoadingContainer>
+              )}
+
+              <Styles.FormContainer isLoading={isLoading}>
```
chrisrhymes, line 187: Hide form with CSS when isLoading instead of removing from the page.

DamienBurkeWN (approved):
Good stuff 👍

PR #156, opened 2022-03-14
Add optional breadcrumb to HeroImage

chrisrhymes:
Added the option to display the parent breadcrumb in the hero image. By default it is disabled. If set to display, it will display the breadcrumb above the title in the hero image. 

As part of this it was requested to make the hero image content the maximum width of the page. This involved more refactoring than I thought and ended up using flexbox for positioning and size, instead of position: absolute. Quite a big change for a small change. 

## To test
- `npm run test HeroImage`
- Check the HeroImage in structure, ensuring it displays as expected in different screen resolutions
- Check the HeroImage in the ServiceLandingPage examples, ensuring it displays as expected in different screen resolutions

### DIFF
--- a/src/library/structure/HeroImage/HeroImage.styles.js
+++ b/src/library/structure/HeroImage/HeroImage.styles.js
@@ -140,10 +140,18 @@ export const CallToActionLink = styled.a`
 export const BreadcrumbLink = styled.a`
   ${(props) => props.theme.fontStyles}
   ${(props) => props.theme.linkStyles}
+  color: ${(props) =>
+    props.backgroundBox ? props.theme.theme_vars.colours.action : props.theme.theme_vars.colours.white};
   margin-bottom: ${(props) => props.theme.theme_vars.spacingSizes.small};
   display: block;
 
-  &:hover {
+  &:hover,
+  &:focus {
     text-decoration-style: dashed;
+    text-shadow: ${(props) =>
+      props.backgroundBox
+        ? 'none'
+        : `2px 2px 4px rgba(150, 150, 150, 0.5), -2px 2px 4px rgba(150, 150, 150, 0.5),
+      2px -2px 4px rgba(150, 150, 150, 0.5), -2px -2px 4px rgba(150, 150, 150, 0.5)`};
   }
 `;
--- a/src/library/structure/HeroImage/HeroImage.styles.js
+++ b/src/library/structure/HeroImage/HeroImage.styles.js
@@ -1,5 +1,5 @@
-import styled from "styled-components";
-import Heading from "../../components/Heading/Heading";
+import styled from 'styled-components';
+import Heading from '../../components/Heading/Heading';
 
 /**
  * Hero image container with optional dark gradient for non-box mode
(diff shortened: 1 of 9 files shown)


src/library/structure/HeroImage/HeroImage.styles.js
```diff
@@ -1,27 +1,44 @@
...
   background-position: center;
   background-repeat: no-repeat;
   background-size: cover;
-  position: relative;
+  display: flex;
+  justify-content: center;
+  flex-direction: row;
+  align-items: flex-end;
+
+  &::before {
```
chrisrhymes, line 22: Use psudo element to control the aspect ratio. This should allow the height to grow with more content https://css-tricks.com/aspect-ratio-boxes/#aa-the-pseudo-element-tactic

src/library/structure/HeroImage/HeroImage.styles.js
```diff
@@ -107,12 +120,21 @@ export const CallToActionLink = styled.a`
...
-      -2px -2px 4px rgba(150, 150, 150, 0.5);
+    text-shadow: 2px 2px 4px rgba(150, 150, 150, 0.5), -2px 2px 4px rgba(150, 150, 150, 0.5),
+      2px -2px 4px rgba(150, 150, 150, 0.5), -2px -2px 4px rgba(150, 150, 150, 0.5);
   }
   &:active {
     transform: translate(3px);
   }
 `;
+
+export const BreadcrumbLink = styled.a`
```
chrisrhymes, line 131: Added the breadcrumb link styles

src/library/structure/HeroImage/HeroImage.types.ts
```diff
@@ -38,4 +40,14 @@ export interface HeroImageProps {
    * Set to true to put the text and call to action on a white box, false to overlay on a darkened image
    */
   backgroundBox: boolean;
+
+  /**
+   * Should the breadcrumb be shown
+   */
+  showBreadcrumb?: boolean;
```
DamienBurkeWN, line 47: Do we need this? Could we instead rely on whether breadcrumbs array is empty or not?

src/library/structure/HeroImage/HeroImage.types.ts
```diff
@@ -38,4 +40,14 @@ export interface HeroImageProps {
...
+
+  /**
+   * An optional array of breadcrumbs
+   */
+  breadcrumbsArray?: BreadcrumbProp[];
```
DamienBurkeWN, line 52: I know it has been done already in some other code but I don't think including the _type_ in the prop name is stylistically good so just **breadcrumbs** would be better. However, in this case, as we only ever look at the last one in the array, why do we even need it to be an array? I feel this component would be better off accepting a single **BreadcrumbProp** called **breadcrumb** (and being set or not handles the job of showBreadcrumb too).

DamienBurkeWN (changes requested):
I'm unclear on the 'make the hero image content the maximum width of the page' aspect - if that means moving the box/title area to align with the page width container, I'm sure there was a discussion about doing that and we agreed not to. The reasoning was that the hero image itself is chosen so that the bottom left is the 'safe' area to obscure with any content. If the box/title moves depending on the browser size, there is no 'safe' area anymore?

Anyway, if that is how it will be, there's some other issues now - on mobile screen sizes, the boxed content area is now no longer in the center (right side is closer to the scrollbar than before) and the call to action button loses its spacing from the bottom of the box.

As it is currently: 
image

Compare now:
image

src/library/structure/HeroImage/HeroImage.types.ts
DamienBurkeWN: No, I'd do that in the frontend, we've got the breadcrumb trail at the page component level, easy enough to pass the last one through to this component. I think we can safely assume it'll never now be used as a slice/embedded para (and we can remove the support for that use in the frontend too).

DamienBurkeWN: (if it was ever to be used as a slice, it surely wouldn't have a breadcrumb on it?)

chrisrhymes:
Thanks for the feedback. The safe area is quite difficult to define, as the image ratio changes on mobile and desktop so you will see different parts of the image. The overlay box provides one solution and there is the linear gradient when the box is not used. I think whatever we do it will need a bit of trial and error by the content editors to work out what works best. 

I have tidied up the mobile and tablet styles so it looks more like before. Please can you retest. 

I have removed the showBreadcrumb prop and updated the `breadcrumbsArray` to be a single `breadcrumb`. If it's not provided then it will not be displayed.

DamienBurkeWN (approved):
Nice 👍

chrisrhymes:
Good spot. I have updated the breadcrumb to match the call to action link styles when there is no background box

DamienBurkeWN (approved):
👍

PR #158, opened 2022-03-18
Add adult social care icons

chrisrhymes:
Added the new icons for adult social care. A backend PR will also be need to be created to allow selection from the CMS.


DamienBurkeWN (approved):
Just the one typo on the icon naming otherwise 👍

PR #175, opened 2022-04-20
Fix SectionLinks alt attribute

chrisrhymes:
Removing the alt text as it is invalid on non img elements. Updated tests to look for aria-label instead of alt attributes. 

Using span instead of figure and following these guidelines https://www.davidmacd.com/blog/alternate-text-for-css-background-images.html


DamienBurkeWN (approved):
👍

PR #191, opened 2022-05-16
Clicking link in alert banner dismisses banner

DamienBurkeWN:
## Testing
* In Storybook you can play with the AlertBanner structure to verify that clicking a link within a banner also dismisses it
* Also fixed the bug whereby a dismissed site alert still flashes up briefly on initial page load in production build; `yalc publish` this branch and on the frontend `yalc add northants-design-system && yarn install && yarn bs`, set a sitewide alert in the CMS, load homepage in frontend, dismiss it, reload page, verify no banner flash.

### DIFF
--- a/src/library/structure/AlertBanner/AlertBanner.stories.tsx
+++ b/src/library/structure/AlertBanner/AlertBanner.stories.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { Story } from '@storybook/react/types-6-0';
 import AlertBanner from './AlertBanner';
 import { AlertBannerProps } from './AlertBanner.types';
+import Button from '../../components/Button/Button';
 import MaxWidthContainer from '../MaxWidthContainer/MaxWidthContainer';
 
 export default {
@@ -20,6 +21,11 @@ const childrenMessage = (
   </p>
 );
 
+const resetAlert = (event) => {
+    window.localStorage.clear();
+    window.location.reload();
+};
+
 const Template: Story<AlertBannerProps> = (args) => (
   <>
     <AlertBanner {...args} />
@@ -28,6 +34,8 @@ const Template: Story<AlertBannerProps> = (args) => (
       <br />
       <br />
       <p>Once this unique alert has been hidden it will not be shown again on this device.</p>
+      <br />
+      <Button url={null} onClick={resetAlert}>Click to reset</Button>
     </MaxWidthContainer>
   </>
 );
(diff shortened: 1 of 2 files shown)


src/library/structure/AlertBanner/AlertBanner.stories.tsx
```diff
@@ -28,6 +34,8 @@ const Template: Story<AlertBannerProps> = (args) => (
       <br />
       <br />
       <p>Once this unique alert has been hidden it will not be shown again on this device.</p>
+      <br />
+      <Button url={null} onClick={resetAlert}>Click to reset</Button>
```
DamienBurkeWN, line 38: Just for our convenience...

src/library/structure/AlertBanner/AlertBanner.stories.tsx
```diff
@@ -20,6 +21,11 @@ const childrenMessage = (
   </p>
```
chrisrhymes, line 21: Is it worth setting the href="#" so it doesn't redirect to the homepage of storybook?

src/library/structure/AlertBanner/AlertBanner.stories.tsx
chrisrhymes: Nice!

chrisrhymes (changes requested):
Something is not quite right. If you go on the positive alert example and then dismiss it, then "Click to reset" it is no longer blue, but has a red background? Not sure if it always did that.

src/library/structure/AlertBanner/AlertBanner.stories.tsx
```diff
@@ -16,10 +17,15 @@ export default {
...
 const childrenMessage = (
   <p>
-    Coronavirus | National lockdown: stay at home. <a href="/">Learn what this means for residents and workers here</a>
+    Coronavirus | National lockdown: stay at home. <a href="javascript:;">Learn what this means for residents and workers here</a>
   </p>
 );
 
+const resetAlert = (event) => {
+    window.localStorage.clear();
+    window.location.replace(window.location.href.split('&')[0]);
```
DamienBurkeWN, line 26: For some reason storybook had rewritten the URL to include empty args, so the straight reload() call ended up with an undefined type for the alert type arg. This strips out any args and I've used replace() so your browser history isn't cluttered with another entry too.

chrisrhymes (approved):
👍

PR #194, opened 2022-05-18
Heading inside div instead of span

chrisrhymes:
HeadingWithIcon was set to be contained in a span to resolve an earlier issue with HTML validation. The homepage ServiceLinksBoxed has been refactored to no longer use the HeadingWithIcon so it can be reverted to div for the container. 

https://validator.w3.org/nu/?doc=https%3A%2F%2Fwww.westnorthants.gov.uk%2Fbusiness-and-licensing 

## Testing
- Checkout this branch and `npm run dev` 
- Validate the HTML in the Service Page Example to test it no longer shows the error

### DIFF
--- a/src/library/components/HeadingWithIcon/HeadingWithIcon.styles.js
+++ b/src/library/components/HeadingWithIcon/HeadingWithIcon.styles.js
@@ -1,6 +1,6 @@
 import styled from 'styled-components';
 
-export const HeadingWrapper = styled.span`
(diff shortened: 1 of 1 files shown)


MeganCrowthers:
Task linked: [CU-2b5q7vj ServiceLandingPage html validation issue](https://app.clickup.com/t/2b5q7vj)

DamienBurkeWN (approved):
Powerful stuff 😆

PR #195, opened 2022-05-19
Fix for search suggestions in the header

chrisrhymes:
- Checkout this branch
- `npm run dev`
- View the Structure/Header component and search for 'co' and the search suggestions should no longer be cropped
- `yalc publish` then in frontend `yalc add northants-design-system && yarn install && yarn dev`
- Test the search suggestions display correctly in the frontend


DamienBurkeWN (approved):
😲 👍

PR #199, opened 2022-05-24
Add card component and cards slice

chrisrhymes:
- New Card component with optional image, header and footer link. The content is required. 
- Use srcset for card images to load smaller images on mobile
- New Cards slice that allows for multiple cards to be presented in a grid, 3 across on desktop, 2 on tablet and single on mobile


MeganCrowthers:
Task linked: [CU-2849g6m 3 column text slice (post-its)](https://app.clickup.com/t/2849g6m)

src/library/components/Card/Card.types.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Optional small image
+   */
+  imageSmall?: string;
+
+  /**
+   * Optional image alt text
+   */
+  imageAltText?: string | null;
```
DamienBurkeWN, line 25: Why the | null, isn't the question mark enough?

src/library/components/Card/Card.types.ts
```diff
@@ -0,0 +1,43 @@
...
+   * Optional card header
+   */
+  header?: string;
+
+  /**
+   * The card content as a string of html
+   */
+  content: string;
+
+  /** Optional large image */
```
DamienBurkeWN, line 14: Formatting different to all the other comments

src/library/components/Card/Card.types.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface CardProps {
+  /**
+   * Optional card header
+   */
+  header?: string;
+
+  /**
+   * The card content as a string of html
+   */
+  content: string;
```
DamienBurkeWN, line 12: Shouldn't this be optional, so that cards can be used as a triple image slice - no content, no heading, no footer, just images set

DamienBurkeWN (changes requested):
Needs a bit of tweaking to better handle being used as triple image slice - currently, content is mandatory, and if you don't set anything except the image, there's a blank white bit at the bottom of the card.

src/library/components/Card/Card.types.ts
chrisrhymes: The frontend was complaining that the types were incompatible with the GraphQL types

PR #202, opened 2022-05-25
Allow Image component to have text wrapped up against it

DamienBurkeWN:
### Testing
* Look at Image stories 
* Look at content page story and compare appearance (some minor list formatting differences will be evident)
* Probably best tested with real content in frontend, PRs to follow


src/library/slices/Image/Image.styles.js
```diff
@@ -8,6 +8,16 @@ export const Container = styled.figure`
       : props.theme.theme_vars.colours.grey_light};
   border-bottom: 1px solid ${(props) => props.theme.theme_vars.colours.grey};
   margin-bottom: 15px;
+  float: ${(props) => (props.wrapText === true ? 'left' : 'none')};
```
DamienBurkeWN, line 11: So basically just drop the image size down and float it to the left if wrapText is used, but only on screens bigger than mobile

src/themes/theme_generator.jsx
```diff
@@ -26,7 +26,7 @@ const generate_theme = (theme_vars) => {
                 line-height: 1.45;
 
                 p, h2, h3, h4, ul, ol {
-                    max-width: 750px;
+                    max-width: 55rem;
```
DamienBurkeWN, line 29: While I was there, the slight increase in available space discussed on Slack

.storybook/preview.js
```diff
@@ -1,10 +1,8 @@
 import React from 'react';
 import { ThemeProvider } from 'styled-components';
-import { addDecorator } from '@storybook/react';
```
DamienBurkeWN, line 3: Wasn't used

.storybook/preview.js
```diff
@@ -1,10 +1,8 @@
...
 import { themes } from '../src/themes/theme_generator';
-import { createGlobalStyle } from 'styled-components';
```
DamienBurkeWN, line 5: No longer necessary

src/library/structure/PageMain/PageMain.styles.js
```diff
@@ -27,35 +27,6 @@ export const Container = styled.main`
         font-style: italic;
     }
 
-    ul, ol {
```
DamienBurkeWN, line 30: Removed all this stuff putting back in margins and a bullet disc at a hard-coded position!

src/themes/GlobalStyleReset.jsx
```diff
@@ -0,0 +1,146 @@
+import { createGlobalStyle } from 'styled-components';
```
DamienBurkeWN, line 1: This is basically reset.css

src/themes/GlobalStyleReset.jsx
```diff
@@ -0,0 +1,146 @@
...
+  menu,
+  nav,
+  section {
+    display: block;
+  }
+  body {
+    line-height: 1;
+  }
+  ol {
+    list-style: decimal inside;
```
DamienBurkeWN, line 124: ...with sensible list defaults

chrisrhymes: The 2rem margin-left feels too much to me, 1 or 1.25 would be better. It's now 32px rather than the previous 20px.

chrisrhymes (changes requested):
Good way of modifying an existing component to work with the content.

My question is does the add value to the page if it's smaller? Wider discussion point I guess.

> Avoid unnecessary decoration. Only use images if there’s a real user need.
https://design-system.service.gov.uk/styles/images/

- I think the margin-left needs to be reduced compared to previous styling
- The [Contents component](https://northants-design-system.netlify.app/?path=/story/page-examples-service-page-example--service-page-with-contents) styling has now changed after the defaults have been changed. 
- The footer links appear out of line with the icons below
- The ServiceLinksList now have two bullets

src/library/structure/Footer/Footer.styles.js
```diff
@@ -1,35 +1,41 @@
...
+
 export const StyledMaxWidthContainer = styled(MaxWidthContainer)`
-  color: ${props => props.theme.theme_vars.colours.white};
-`
+  color: ${(props) => props.theme.theme_vars.colours.white};
+`;
+
 export const FooterList = styled.ul`
   padding: 0px;
+  margin-left: 0;
```
DamienBurkeWN, line 15: Fix for alignment shift

src/library/structure/ServicesLinksList/ServicesLinksList.styles.js
```diff
@@ -170,6 +171,7 @@ export const QuicklinkList = styled.ul`
   display: none;
   margin: 0 !important;
   padding: 0 !important;
+  list-style-type: none;
```
DamienBurkeWN, line 174: Fix for double discs

DamienBurkeWN:
Does it add value... well that's for the marketing boys. I don't think it will be used on purely informational pages, more for the flashier marketing driven campaign style pages.

src/library/structure/ContentsNavigation/ContentsNavigation.stories.tsx
```diff
@@ -3,7 +3,7 @@ import { Story } from '@storybook/react/types-6-0';
 import ContentsNavigation from './ContentsNavigation';
 import { ContentsProps } from '../Contents/Contents.types';
 import { SBPadding } from '../../../../.storybook/SBPadding';
-import { ContentsArray } from '../Contents/Contents.storydata';
```
DamienBurkeWN, line 6: Bonus bugfix

chrisrhymes (changes requested):
If the list item contains long text then subsequent lines aren't indented to where the first line text starts. This doesn't look right to me. 
Screenshot 2022-05-26 at 16 59 11

chrisrhymes (approved):
Gotta love CSS!
image

PR #219, opened 2022-06-13
Add bundle-analyzer to rollup

chrisrhymes:
Not sure if we need to include this in the actual design system, but found it useful to show the largest files exported from our design system so we can discuss what to improve. 

Running `npm run build` will provide analysis. For more details, change the `summaryOnly` to false in the rollup.config.js file.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -4,6 +4,7 @@ import commonjs from '@rollup/plugin-commonjs';
 import typescript from 'rollup-plugin-typescript2';
 import svg from 'rollup-plugin-svg';
 import json from '@rollup/plugin-json';
+import analyze from 'rollup-plugin-analyzer';
 
 const packageJson = require('./package.json');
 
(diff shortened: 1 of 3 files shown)


rollup.config.js
```diff
@@ -27,12 +28,10 @@ export default {
     peerDepsExternal(),
     resolve({
       extensions: extensions,
+      preferBuiltins: true,
```
chrisrhymes, line 31: This removes the warnings when running `npm run build`

DamienBurkeWN:
Like this. It makes zero difference to the resulting build files so it isn't loading any extra stuff in the actual build.

chrisrhymes:
We can further reduce the bundle size by changing some of the dependencies to dev dependencies and then adding them to the peer dependencies. This would work for axios and downshift, which are quite large packages. 

The issue doesn't really go away though as you have to then add them as dependencies on the frontend, so the frontend build size is the same.

DamienBurkeWN:
I don't think it serves any purpose to move axios & downshift, as you say same end result in terms of size.

chrisrhymes:
Stale PR. Fixes for Rollup config have been implemented in #349

PR #220, opened 2022-06-14
Add url and cookie helper tests

chrisrhymes:
Added tests for the cookie and url helpers so that we can document what they do in case we need to refactor them in future. 

I put the tests in the `__tests__` directory as I didn't want to clutter up the `helpers` directory and it gives us a location to write other non component tests in future.

### DIFF
--- a/__tests__/UrlHelpers.test.ts
+++ b/__tests__/UrlHelpers.test.ts
@@ -1,0 +1,157 @@
+import {
+  handleParams,
+  removeParams,
+  removeValueFromParam,
+  countParams,
+  getParamValue,
+  getDropDownValues,
+  getCheckboxValues,
+  deSlug,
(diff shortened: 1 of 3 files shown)


__tests__/UrlHelpers.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    it('removes the name parameter', () => {
+      removeParams(['name']);
+
+      expect(window.location.href).toEqual('http://dummy.com/?page=1');
+    });
+  });
+
+  /**
+   * Uses ASCI code `%2C` instead of `,` to separate multiple params
```
DamienBurkeWN, line 90: ASCII 😄  And strictly speaking it's URL encoded ASCII.

DamienBurkeWN (approved):
Minor comment nitpick if you're bored enough to change it

PR #227, opened 2022-06-22
Prevent cropping of link text in services links list

chrisrhymes:
The default list style `left: 1.25rem;` was pushing the list item content outside the container in certain circumstances. This resets back to zero to prevent it. 

To test, checkout this branch and `npm run dev` then visit the Home Page example and sort the services alphabetically. The 'Report a highways problem' text should no longer be cropped.


DamienBurkeWN (approved):
👍

PR #235, opened 2022-07-05
Remove image wrapText feature

chrisrhymes:
Remove the wrapText feature from Image as this is now handled by the ImageAndText slice. Also handle if ImageAndText doesn't have an image set. 

- https://github.com/FutureNorthants/northants-website/pull/512#pullrequestreview-1027483428


DamienBurkeWN (approved):
👍

PR #245, opened 2022-07-19
Add bins and waste icons

chrisrhymes:


DamienBurkeWN (approved):
These are great, particularly like the fly tipping and food waste ones!

PR #246, opened 2022-07-20
Display service icons in Iconography docs

chrisrhymes:
- Remove the icon stories and display them in the [iconography docs page](https://deploy-preview-246--northants-design-system.netlify.app/?path=/docs/documentation-iconography--all-service-icons)
- Refactor the service icons so that they have consistent directory naming and the Icon and IconHover are in the same directory (for example PlanningIcon/Icon -> Planning/Icon)
- Update imports in ServiceIcons.ts


DamienBurkeWN:
Brilliant! Make it so!

MeganCrowthers:
Task linked: [CU-2pe5pbf Icons documentation refactoring](https://app.clickup.com/t/2pe5pbf)

src/docs/4Iconography.stories.mdx
```diff
@@ -1,20 +1,112 @@
...
-import CouncilTaxIcon from '../library/components/icons/services/CouncilTaxIcon/Icon';
-import CovidIcon from '../library/components/icons/services/CovidIcon/Icon';
-
-<Meta title="Documentation/Iconography" />
+<Meta title="Documentation/Iconography" component={HeadingWithIcon} />
 
 # Iconography
 
-text here
+There are three main types of icons used in the design system.
```
DamienBurkeWN, line 24: Not sure why but all this text does not display for me unless I toggle to the docs panel. All the other docs pages display correctly in both canvas and docs panels?

DamienBurkeWN (changes requested):

src/docs/4Iconography.stories.mdx
chrisrhymes: It seems this is the way storybook is designed to work. 

If you include `<Story>` in your mdx file then the text around stories will only appear in the docs tab. 

> To get a "documentation-only story" in your UI, define a <Meta> as you usually would, but don't define any stories. It will show up in your UI as a documentation node
- https://storybook.js.org/docs/react/api/mdx#documentation-only-mdx

DamienBurkeWN (approved):
👍

PR #252, opened 2022-08-05
Prevent scrollbar with nested list in table

chrisrhymes:
Issue occurs in revision for 22/07 2022 09:25
https://staging.westnorthants.gov.uk/visiting-library/computers-internet-and-wifi 

This fix removes the default left on list items when inside a `.table-container`. 

## Testing
- Checkout this branch and comment out the css change in `GlobalStyleReset.jsx`
- Run `npm run dev` and view the Content Page example. Scroll down to the 'Example table with a nested list' and there should be a scrollbar
- Uncomment the css change and save the file
- Refresh the Content Page example and the scrollbar should be removed

## Frontend testing
- `yalc publish` and then in the frontend `yalc add northants-design-system && yarn install && yarn dev` then visit a page with a list in a table and the scrollbar should be removed

### DIFF
--- a/src/library/pages/ContentPage/ContentPage.tsx
+++ b/src/library/pages/ContentPage/ContentPage.tsx
@@ -143,8 +143,8 @@ export const ContentPage: React.FunctionComponent<ContentPageProps> = ({}) => (
               </tr>
             </thead>
             <tbody>
-              {smallTable.data.map((row) => (
-                <tr key={row}>
+              {smallTable.data.map((row, index) => (
+                <tr key={index}>
                   {row.map((item) => (
                     <td key={item}>{item}</td>
                   ))}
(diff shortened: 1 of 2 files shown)


src/library/pages/ContentPage/ContentPage.tsx
```diff
@@ -143,8 +143,8 @@ export const ContentPage: React.FunctionComponent<ContentPageProps> = ({}) => (
               </tr>
             </thead>
             <tbody>
-              {smallTable.data.map((row) => (
-                <tr key={row}>
+              {smallTable.data.map((row, index) => (
+                <tr key={index}>
```
chrisrhymes, line 147: Fix for incorrect key

DamienBurkeWN (approved):
👍

PR #260, opened 2022-08-16
Fix/row min width

chrisrhymes:
This affected NewsArticleFeaturedBlock when the title content wasn't wide enough to push the Row wider than the min-width. 

I've shortened the titles in the story data to help replicate the issue.

## Testing
- Test NewsArticleFeaturedBlock component with 3 articles story
- With the Row fix the `ul` should be the full width of the containing div
- Revert the Row min-width css fix and the `ul` is no longer as wide as the containing div

### DIFF
--- a/src/library/structure/NewsArticleFeaturedBlock/NewsArticleFeaturedBlock.stories.tsx
+++ b/src/library/structure/NewsArticleFeaturedBlock/NewsArticleFeaturedBlock.stories.tsx
@@ -1,46 +1,54 @@
-import React from "react";
+import React from 'react';
 import { Story } from '@storybook/react/types-6-0';
-import NewsArticleFeaturedBlock from "./NewsArticleFeaturedBlock";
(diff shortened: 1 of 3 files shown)


src/library/structure/NewsArticleFeaturedBlock/NewsArticleFeaturedBlock.stories.tsx
```diff
@@ -1,46 +1,55 @@
...
+export const ExampleNoTitleOrButton = Template.bind({});
 ExampleNoTitleOrButton.args = {
   articles: NewsArticleData,
   withoutTitle: true,
-  viewAllLink: ""
+  viewAllLink: '',
 };
 
-export const Example2Articles = Template.bind({});    
+export const Example3Articles = Template.bind({});
```
chrisrhymes, line 42: New story, the rest is prettier

PR #264, opened 2022-08-22
Update test runner versions

chrisrhymes:
Ubuntu 18 test runners are being deprecated so swapping to Ubuntu version 22.04. 
https://github.com/actions/runner-images/issues/6002

Also updating to use the new versions of checkout and setup-node actions.


DamienBurkeWN (approved):
👍

PR #266, opened 2022-08-24
Fix label and hint hiding. Add test

chrisrhymes:
The label and hint were set to 'string' type and the code checked if they were null to set whether they should be hidden. 

- I updated the types to allow null and then updated the styles to use `display: none;` instead of visually hidden. 
- Added a test to check that they are hidden when they are supposed to be. 
- Added testid to hint and legend as I wasn't sure how to find them with tests when they have no text content.
- Updated story to show label and hint, then added a new story to show them hidden

### DIFF
--- a/src/library/components/CheckboxListFilter/CheckboxListFilter.styles.js
+++ b/src/library/components/CheckboxListFilter/CheckboxListFilter.styles.js
@@ -1,4 +1,5 @@
 import styled, { css } from 'styled-components';
+import { VisuallyHidden } from './../../helpers/style-helpers';
 
 export const Container = styled.div`
   ${(props) => props.theme.fontStyles};
@@ -19,14 +20,8 @@ export const Fieldset = styled.fieldset`
 `;
 
 const hideLabel = (props) => {
-  if (props.labelHidden === true) {
-    return css`
-      display: none;
-    `;
-  } else {
-    return css`
-      display: table;
-    `;
+  if (props.labelHidden == true) {
+    return VisuallyHidden;
   }
 };
 
@@ -34,6 +29,7 @@ export const Legend = styled.legend`
   color: #0b0c0c;
   -webkit-box-sizing: border-box;
   box-sizing: border-box;
+  display: table;
   max-width: 100%;
   margin-bottom: 10px;
   padding: 0;
--- a/src/library/components/CheckboxListFilter/CheckboxListFilter.styles.js
+++ b/src/library/components/CheckboxListFilter/CheckboxListFilter.styles.js
@@ -20,7 +20,7 @@ export const Fieldset = styled.fieldset`
 `;
 
 const hideLabel = (props) => {
-  if (props.labelHidden == true) {
+  if (props.labelHidden) {
     return VisuallyHidden;
   }
 };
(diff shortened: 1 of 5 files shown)


MeganCrowthers:
Task linked: [CU-2rd3txw CheckboxListFilter component review](https://app.clickup.com/t/2rd3txw)

src/library/components/CheckboxListFilter/CheckboxListFilter.styles.js
```diff
@@ -20,7 +20,7 @@ export const Fieldset = styled.fieldset`
 `;
 
 const hideLabel = (props) => {
-  if (props.labelHidden === true) {
+  if (props.labelHidden == true) {
```
DamienBurkeWN, line 23: No need for the == true then?

src/library/components/CheckboxListFilter/CheckboxListFilter.styles.js
```diff
@@ -38,14 +38,19 @@ export const Legend = styled.legend`
 `;
 
 const hideHint = (props) => {
-  if (props.hintHidden === true) {
-    return VisuallyHidden;
+  if (props.hintHidden) {
```
DamienBurkeWN, line 41: Doesn't this need reverting too?

DamienBurkeWN (changes requested):
Looking at it, I think you could have gotten away with display none for these as there's genuinely no content when null? Anyway, it's in a mixed state at the moment with the hint treated differently to the label.

src/library/components/CheckboxListFilter/CheckboxListFilter.tsx
```diff
@@ -44,8 +44,10 @@ const CheckboxListFilter: React.FunctionComponent<CheckboxListFilterProps> = ({
   return (
     <Styles.Container data-testid="CheckboxListFilter">
       <Styles.Fieldset aria-describedby="waste-hint">
-        <Styles.Legend labelHidden={labelHidden}>{label}</Styles.Legend>
-        <Styles.Hint id="waste-hint" hintHidden={hintHidden}>
+        <Styles.Legend labelHidden={labelHidden} data-testid="CheckboxListFilterLegend">
+          {label}
+        </Styles.Legend>
+        <Styles.Hint id="waste-hint" hintHidden={hintHidden} data-testid="CheckboxListFilterHint">
```
chrisrhymes, line 50: I was wondering if we should prevent the Styles.Hint from rendering at all if the hint is null?  The fieldset has aria-describedby the hint, but the hint is generally empty in our usage situation.  

I'm also confused why the id is hardcoded as "waste-hint". If the component is used multiple times on a page it would duplicate the same id?

DamienBurkeWN: Hadn't spotted the hardcoded id, yeah that's a bit rubbish...

chrisrhymes: Looks like it was copied from the example html on gov.uk https://design-system.service.gov.uk/components/checkboxes/

DamienBurkeWN: 🤣 quality

DamienBurkeWN (approved):

PR #267, opened 2022-08-24
All Services link should be isList

chrisrhymes:
Without the `isList` it renders as a `<div>` instead of an `<li>` inside the parent Row which also has `isLists` set which converts it to an `<ul>`. 

The other links are set correctly in the map(). 

See error 9 onwards for details:
https://validator.w3.org/nu/?doc=https%3A%2F%2Fwww.westnorthants.gov.uk%2F


MeganCrowthers:
Task linked: [CU-2rjx257 ServiceLinksBoxed html validation error](https://app.clickup.com/t/2rjx257)

DamienBurkeWN (approved):
👍

PR #283, opened 2022-09-23
Add trailing slash to instagram urls

chrisrhymes:
Prevent sitemorse detecting the links as redirects, as instagram redirects to the address with a trailing slash.


DamienBurkeWN (approved):
Incredible work 👏 🥇

PR #308, opened 2022-12-08
Feature/maps

chrisrhymes:
This PR adds 3 new components:
- Map
- MapMarker
- StaticMap

These have been added as a requirement for the Local Offer project, but I thought it would be easier to review and test as a separate pull request. 

The Map uses the [Google Maps JavaScript API](https://developers.google.com/maps/documentation/javascript) via the `@googlemaps/react-wrapper`. The MapMarker should be placed as a direct child of the Map. 

The StaticMap uses the [Google Maps Static API](https://developers.google.com/maps/documentation/maps-static/overview) and returns an image. This is useful for when users have not accepted cookies. 

## Testing
- Add `NEXT_PUBLIC_GOOGLE_MAPS_API_KEY=` to your env file, and add the api key
- Checkout this branch and run `npm install` then `npm run dev` to test out the components


src/library/components/MapMarker/MapMarker.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { ThemeContext } from 'styled-components';
+
+/**
+ * A Google Maps marker to be placed inside a Map component
+ */
+const MapMarker: React.FunctionComponent<google.maps.MarkerOptions> = (options) => {
+  const [marker, setMarker] = React.useState<google.maps.Marker>();
+  const themeContext = useContext(ThemeContext);
+
+  const icon: google.maps.Symbol = {
```
chrisrhymes, line 11: This is a custom SVG so we can change the appearance and fill colour based on the theme. The anchor and label origin define the alignment for the marker icon and the label that sits on top of the icon.

src/library/components/MapMarker/MapMarker.tsx
```diff
@@ -0,0 +1,63 @@
...
+    path: 'M19.7,5.6c-0.1,0-0.5,0.1-0.9,0.1c-3.3,0.4-6.5,2-8.9,4.4c-2.6,2.6-4.2,5.9-4.6,9.6c-0.1,1,0,3.1,0.1,4.1c0.3,2.1,1.1,4.2,2.3,6.6c0.9,1.6,1.6,2.9,4.2,7c0.9,1.4,1.9,3.1,2.6,4.2c2.3,3.9,4.1,7.9,5.4,11.9c0.2,0.6,0.3,0.8,0.6,0.9c0.5,0.2,1.1,0.1,1.4-0.3c0.1-0.1,0.2-0.5,0.4-0.9c1.1-3.3,2.5-6.5,4.2-9.5c1.2-2.2,2-3.6,4.4-7.4c1.7-2.7,2.3-3.6,2.9-4.8c1.6-2.9,2.5-5.3,2.9-7.8c0.2-1,0.2-3.1,0.1-4.1c-0.2-1.5-0.5-3-1-4.2c-2.1-5.2-6.8-8.9-12.4-9.8C22.9,5.6,20.2,5.5,19.7,5.6z',
+    fillColor: `${themeContext.theme_vars.colours.action_dark}`,
+    fillOpacity: 1,
+    anchor: new google.maps.Point(40, 60),
+    labelOrigin: new google.maps.Point(21, 25),
+  };
+
+  const mergedOptions: google.maps.MarkerOptions = {
+    ...options,
+    anchorPoint: new google.maps.Point(-5, -40),
```
chrisrhymes, line 21: anchorPoint is alignment for the infoWindow, relative to the custom icon.

DamienBurkeWN (approved):
Really nice 👍 

Couple of things, not affecting approval.

1. Changing zoom level in the example map doesn't update the map itself unless you click onto another story and then back in. I don't think this would be an issue 'in the wild' unless you provided zoom buttons outside of the map window though.
2. On multi marker example, clicking marker brings up persistent popup. I kinda expected the first to disappear when I clicked on another? Could be an optional kind of behaviour perhaps?

PR #309, opened 2022-12-13
Hero image call to action secondary

chrisrhymes:
Allow the call to action in HeroImage and the link in SearchBox to be marked as secondary. This will show the differentiation between the priority action (searching) and the secondary action (clicking the link). 

The default is set to true for primary as to not effect current usage of the HeroImage.


MeganCrowthers:
Task linked: [CU-39jar6b Call to action - secondary button](https://app.clickup.com/t/39jar6b)

DamienBurkeWN (approved):

PR #317, opened 2023-01-04
Update to React 18

chrisrhymes:
Update design system to use React 18 and update other dependencies to match. 

Main changes involve adding `children` to types, as well as onSubmit and onChange needing to accept a function. 

There are some deprecation warnings when running `npm run dev` but I searched through the codebase for them and they seem to be in node_modules. 

I tried updating axios to v1. It worked in the design system but produced an error in the frontend website (Z_SYNC_FLUSH undefined???) so it has been reverted for now. 

## Testing
- Checkout this branch and `npm install`
- Test build `yalc publish`
- Run tests `npm run test`
- Manually test design system `npm run dev`


.storybook/main.js
```diff
@@ -51,7 +59,9 @@ module.exports = {
       },
     });
     config.resolve.extensions.push('.ts', '.tsx');
-
     return config;
   },
+  docs: {
+    autodocs: true,
```
chrisrhymes, line 65: This is from `npx storybook@next automigrate` which says this will be the default in the next major version of storybook.

package.json
```diff
@@ -82,21 +82,22 @@
     "eslint-plugin-react-hooks": "^4.3.0",
     "eslint-plugin-storybook": "^0.6.8",
     "identity-obj-proxy": "^3.0.0",
-    "jest": "^27",
+    "jest": "^29",
+    "jest-environment-jsdom": "^29.3.1",
```
chrisrhymes, line 86: jest-environment-jsdom is not installed as part of jest in the latest version so is needed for tests to work

src/library/components/Button/Button.types.ts
```diff
@@ -1,38 +1,53 @@
...
   /**
    * Optional override for the colour
    */
-  colourOverride?: string; 
+  colourOverride?: string;
+
+  /**
+   * Button contents
+   */
+  children?: React.ReactNode;
```
chrisrhymes, line 52: children is now required in type definitions

src/library/components/PostCodeAddressTable/PostCodeAddressTable.tsx
```diff
@@ -29,8 +29,8 @@ const PostCodeAddressTable: React.FunctionComponent<PostCodeAddressTableProps> =
                 {results.addresses.map((addr, i) => (
                   <tr key={i}>
                     <th scope="row">{addr.single_line_address}</th>
-                    <td>{addr.ctaxband}</td>
```
chrisrhymes, line 32: This was flagged during `yalc publish`. The band and amount is not in the API response. Need to discuss if this component is still required.

src/library/pages/Home/HomePage.tsx
```diff
@@ -36,35 +36,7 @@ export const HomePage: React.FunctionComponent<HomePageProps> = ({
...
-              hjid: 12345,
-              hjsv: 6,
-            };
-            a = o.getElementsByTagName('head')[0];
-            r = o.createElement('script');
-            r.async = 1;
-            r.src = t + h._hjSettings.hjid + j + h._hjSettings.hjsv;
-            a.appendChild(r);
-          })(window, document, 'https://static.hotjar.com/c/hotjar-', '.js?sv=');
-        }}
```
chrisrhymes, line 67: This script caused errors in `yalc publish` so it has been removed. Same in MemorialPage.

chrisrhymes:
Managed to fix the netlify build by setting an environment variable `NODE_VERSION` to `14` as Next js v13 no longer supports node 12.

DamienBurkeWN (approved):
Looks OK - suggest checking with Richard re Hotjar removal

MeganCrowthers:
Task linked: [CU-2yfr594 Update design system and frontend to React 18](https://app.clickup.com/t/2yfr594)

chrisrhymes:
@DamienBurkeWN I think that the pages in the design system are just mock ups and not used in the frontend, so it should be ok to remove it from here.

PR #329, opened 2023-02-01
Inquest Schedule slice

chrisrhymes:
New slice to allow displaying of inquests data from the coroners api. 

It has been asked to make the slice a simple table as per the [existing page](https://www.northamptonshire.gov.uk/councilservices/births-deaths-ceremonies/deaths/Pages/coroners-inquests.aspx).

### DIFF
--- a/src/library/slices/InquestSchedule/InquestSchedule.tsx
+++ b/src/library/slices/InquestSchedule/InquestSchedule.tsx
@@ -1,9 +1,11 @@
 import React from 'react';
-
 import { InquestScheduleProps } from './InquestSchedule.types';
 import * as Styles from './InquestSchedule.styles';
 
-const InquestSchedule: React.FC<InquestScheduleProps> = ({ CaseAppointments, title }) => (
+/**
+ * A table displaying a schedule of inquests
+ */
+const InquestSchedule: React.FunctionComponent<InquestScheduleProps> = ({ caseAppointments, title }) => (
   <Styles.Container data-testid="InquestSchedule">
     <div className="table-container">
       <table>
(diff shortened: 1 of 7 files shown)


src/library/slices/InquestSchedule/InquestSchedule.tsx
```diff
@@ -0,0 +1,54 @@
...
+              <td>{item.placeOfDeath}</td>
+              <td>{new Date(item.dateTimeOfDeath).toLocaleString('en-GB')}</td>
+              <td>{item.coroner}</td>
+              <td>{item.courtroomFullAddress}</td>
+              <td>{new Date(item.startDateTime).toLocaleString('en-GB')}</td>
+            </tr>
+          ))}
+          {caseAppointments.length === 0 && (
+            <tr>
+              <td colSpan={7}>
```
DamienBurkeWN, line 38: I can see your intention is to span across the table width, but the paragraph is limited to a max of 750px so this message... doesn't span across the table width.

src/library/slices/InquestSchedule/InquestSchedule.tsx
```diff
@@ -0,0 +1,54 @@
...
+                <p>We can't find any results at the moment.</p>
+                <p>
+                  This could either be because there are no inquests scheduled for next month or the system is currently
```
DamienBurkeWN, line 41: Do we not know the difference between these two states? What good does retrying later do me if there are none scheduled? Grammatically 'or that the system' would be an improvement too.

DamienBurkeWN (changes requested):
Tweak needed to the no results state I think?

DamienBurkeWN (approved):
👍

PR #331, opened 2023-02-08
Resolve video slice server side render hydration error

chrisrhymes:
Whilst testing something else, I noticed that video slices were causing a hydration error due to the content changing between server side render and client side render. This PR uses useEffect to rerender the content on the client side as it needs access to local storage for the cookies check. 

## Testing 
- Edit a page in the CMS and add a Video slice
- Using preprod_frontend branch, run `yarn dev` and see a hydration error message
- Checkout this branch and run `yalc publish`
- In the frontend run `yalc add northants-design-system` and then `yarn install && yarn dev`
- Visit the same page with the video and the hydration warning should now be gone


DamienBurkeWN (approved):
👍

PR #335, opened 2023-02-20
Make file download title paragraph tag

### LINKED ISSUE #334: File Download missing spaces
Testing tool picking up that there is no space between the last word of the actual download and the file format which is on the line beneath, so for instance on page https://www.westnorthants.gov.uk/schools-and-education/information-schools-staff/school-emergency-plans-and-procedures. 
Although it looks fine to anyone using the website, it may cause a problem to someone using a screenreader

chrisrhymes:
Fixes #334 

Making the title a paragraph tag should mean that screen readers see the file title as a separate piece of content to the file type. 

## Testing
- Checkout this branch and run `npm run dev` testing the FileDownload looks as expected
- Run `yalc publish` and then in the frontend `yalc add northants-design-system` then `yarn install` and `yarn dev`
- Visit a page with file downloads and check that it all appears the same as before, but the title should now be in a paragraph tag.

### DIFF
--- a/src/library/components/FileDownload/FileDownload.stories.tsx
+++ b/src/library/components/FileDownload/FileDownload.stories.tsx
@@ -3,6 +3,8 @@ import { Story } from '@storybook/react/types-6-0';
 import FileDownload from './FileDownload';
 import { FileDownloadProps } from './FileDownload.types';
 import { SBPadding } from '../../../../.storybook/SBPadding';
+import MaxWidthContainer from '../../structure/MaxWidthContainer/MaxWidthContainer';
+import PageMain from '../../structure/PageMain/PageMain';
 
 export default {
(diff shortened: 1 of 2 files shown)


src/library/components/FileDownload/FileDownload.stories.tsx
```diff
@@ -16,7 +18,11 @@ export default {
 
 const Template: Story<FileDownloadProps> = (args) => (
   <SBPadding>
-    <FileDownload {...args} />
+    <MaxWidthContainer>
```
chrisrhymes, line 21: Added MaxWidthContainer and PageMain to replicate the paragraph styles that are on the frontend.

DamienBurkeWN (approved):
👍

PR #338, opened 2023-02-27
Update Council Tax directory slice to use bandings

chrisrhymes:
Fixes https://github.com/FutureNorthants/northants-website/issues/787

Update the Council Tax Directory slice to use the North/West Bandings API endpoints instead of the Parishes endpoints. The bandings endpoint has a different response type to the Parishes endpoint. 

I have left the Parishes types in the design system in case they are needed at a later date. 

## Testing
- Checkout this branch and run `npm run test`
- Run `npm run dev` to manually test the Council Tax Directory slice
- Check that the BandingAPIResponse interface matches the expected response from the live API endpoint.


DamienBurkeWN (approved):
👍

PR #349, opened 2023-03-08
Update storybook to 6.5.16

chrisrhymes:
Updated storybook to the latest version and other dependencies using `npm upgrade` except axios, downshift and rollup. I tried to update rollup to v3 but it caused an exception with the frontend. 
 
Update html-react-parser to v3 (from v1). This drops support for IE, which is no longer required as we are using Next.js 13 on the frontend which no longer supports IE anyway. 

For the frontend to work with v3 of html-react-parser we need to also update domhandler to v5, and we need to update TextWithSlices/index.tsx to import Element from html-react-parser instead of domhandler. 

Also fixed bug with accordion where the server-side id didn't match the client-side id. Updated to use the new useId() hook. 

## Testing
- Checkout this branch and `npm install` then `npm run test` to run tests
- Run `npm run dev` to manually test the design system


rollup.config.mjs
```diff
@@ -4,8 +4,12 @@ import commonjs from '@rollup/plugin-commonjs';
 import typescript from 'rollup-plugin-typescript2';
```
chrisrhymes, line 4: File had to be renamed to .mjs so that rollup treated it as a module instead of common js. This also meant you can no longer use require().

src/library/slices/Accordion/Accordion.tsx
```diff
@@ -42,7 +41,7 @@ const Accordion: React.FunctionComponent<AccordionProps> = ({ sections, isFilter
     setAccordionStates(newStatus);
   };
 
-  const accordionId = `accordion${uniqueID()}`;
+  const accordionId = `accordion${useId()}`;
```
chrisrhymes, line 44: Fixes accordion hydration on the server and client so the id matches.

src/library/components/Card/Card.test.tsx
```diff
@@ -58,11 +58,10 @@ describe('Test Component', () => {
     props.header = undefined;
     props.content = undefined;
 
-    const { getByRole } = renderComponent();
-    const image = getByRole('img');
+    const { getByAltText } = renderComponent();
```
chrisrhymes, line 61: It seems it no longer finds images using getByRole('img') so using the alt text to select the image now.

src/library/components/PostCodeSearch/PostCodeSearch.test.tsx
```diff
@@ -205,7 +205,7 @@ describe('PostCodeSearch', () => {
       expect(component).toHaveTextContent('Select an address to continue');
     });
 
-    fireEvent.change(getByRole('combobox'), { target: { value: '12345678911' } });
+    fireEvent.change(getByLabelText('Select your address'), { target: { value: '12345678911' } });
```
chrisrhymes, line 208: Again, combobox no longer seems to work so using the label text to select the select input now.

chrisrhymes:
Superseded by #388

PR #359, opened 2023-05-04
Add missing scope to inquest schedule table

chrisrhymes:
The inquest schedule slice was missing `scope="col"` in all of the table headings. Learned something new. 

> The scope attribute specifies whether a header cell is a header for a column, row, or group of columns or rows.

https://www.w3schools.com/tags/att_th_scope.asp

## Testing
- Go to the Inquest Schedule slice and the all th should now have the correct scope


DamienBurkeWN (approved):
👍 
Though I thought scope of col was assumed if there were no <td> in the same row, and scope of row was assumed if there were and <th> was the first col...

PR #361, opened 2023-08-16
Add a clear filter button to clear all options for the filter

chrisrhymes:
Add a clear filter button to the DirectoryServiceList accordion items. This unchecks all of the checkboxes or radio buttons for that category. 

## Testing
- Checkout this branch and run `npm run dev` and then visit the DirectoryServiceList component in Directory folder
- Click a few options in the filters and then press `Clear filter` and it should uncheck all of the options within that section. Other sections should remain unchanged
- Publish the design system using`yalc publish` and then pull it into the frontend using `yalc add northants-design-system && yarn && yarn dev`
- Test the local offer page on the frontend, selecting a few filters and then pressing the `Clear filter` button ensuring the radio/checkboxes are unchecked, the url is updated and the results are updated.

### DIFF
--- a/src/library/components/RadioCheckboxInput/RadioCheckboxInput.tsx
+++ b/src/library/components/RadioCheckboxInput/RadioCheckboxInput.tsx
@@ -20,7 +20,7 @@ const RadioCheckboxInput: React.FunctionComponent<RadioCheckboxInputProps> = ({
           value={value}
           id={id}
           name={name}
-          onChange={onChange}
+          onClick={onChange}
           defaultChecked={checked}
         />
         <Styles.CategoryInputLabel isChecked={checked} htmlFor={id} singleSelection={singleSelection}>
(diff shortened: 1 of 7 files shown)


src/library/components/RadioCheckboxInput/RadioCheckboxInput.tsx
```diff
@@ -20,7 +20,7 @@ const RadioCheckboxInput: React.FunctionComponent<RadioCheckboxInputProps> = ({
           value={value}
           id={id}
           name={name}
-          onChange={onChange}
+          onClick={onChange}
```
chrisrhymes, line 23: The radio buttons cleared as expected, but when you click on the radio button that was previously selected there was no events being fired. This was fixed by using onClick instead of onChange.

DamienBurkeWN (approved):
Works a treat 👍

PR #362, opened 2023-08-16
Fix/dlo 155 rich text accreditations

chrisrhymes:
The accreditations field is [being updated to a HTML field](https://github.com/FutureNorthants/OpenReferral/pull/180) to allow more information to be recorded with better formatting, such as a link to the ofsted report. This change sanitizes the HTML and only allows specific tags to be used to prevent the page from breaking on the frontend. 

## Testing
- Checkout this branch and run `npm run dev` to test the Service page example
- Run the tests with `npm run test`
- Use `yalc publish` and then in the frontend use `yalc add northants-design-system && yarn install && yarn dev` then test out service pages still show accreditations as expected.


DamienBurkeWN (approved):
👍

PR #364, opened 2023-08-21
Fix how to contact section showing when no contacts

chrisrhymes:
- Update check to see if the contacts has a length greater than zero before showing.
- Update the story with more realistic values that the API will return to ensure the how to contact section is still hidden

## Testing
- Checkout this branch and run `npm run test` to run the tests
- Run `npm run dev` and view Example Directory Service No Contact Details story to ensure the How to contact section is not show.


DamienBurkeWN (approved):
👍

PR #366, opened 2023-08-22
Remove the message when service is not visitable

chrisrhymes:
The wording in the message does not apply to all services locations that are not visitable. The simplest solution for this defects sprint is to remove the message. 

## Testing
- View the Example Directory Service Not Visitable and check the address is at the bottom of the page and that there is no warning text message next to the address.

### DIFF
--- a/src/library/directory/DirectoryService/DirectoryService.tsx
+++ b/src/library/directory/DirectoryService/DirectoryService.tsx
@@ -247,7 +247,7 @@ const DirectoryService: React.FunctionComponent<DirectoryServiceProps> = ({
                 return location.is_visitable == false;
               })
               .map((location) => (
-                <Row>
+                <Row key={location.id}>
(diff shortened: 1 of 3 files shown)


src/library/directory/DirectoryService/DirectoryService.tsx
```diff
@@ -247,7 +247,7 @@ const DirectoryService: React.FunctionComponent<DirectoryServiceProps> = ({
                 return location.is_visitable == false;
               })
               .map((location) => (
-                <Row>
+                <Row key={location.id}>
```
chrisrhymes, line 250: The new test flagged that the key was not being set in the loop.

DamienBurkeWN (approved):
Suspect this will end up with people trying to visit non-visitable locations but works as advertised 👍

PR #369, opened 2023-08-24
Single taxonomies in a paragraph, multiple in a list

chrisrhymes:
Resolves DLO-172.

Update the service details so when there is only a single value it is in a paragraph tag, otherwise it will put them in a list. 

## Testing
- Checkout this branch and then `npm run dev`
- View the Directory Service story and check that the single value taxonomies are in a paragraph and the multiple in a list


DamienBurkeWN (approved):
👍

PR #371, opened 2023-08-25
Add pagination to the top of DirectoryServiceList

chrisrhymes:
Duplicate the pagination at the top of the DirectoryServiceList so users can navigate through the pages without scrolling all the way to the bottom of the results. 

Also moved the bottom pagination section inside the same isLoading check as the services to simplify. 

## Testing
- Visit the Directory Service List component and see the navigation now at the top and the bottom of the page


DamienBurkeWN (approved):
👍

PR #374, opened 2023-08-31
Show age on service details page

chrisrhymes:
- Display the age on the service details page
- Format the ages using a transform in both the service list and service details
- Update formatting for single items in the summary list so they are in a span instead of paragraph. This removes the bottom margin and helps vertical alignment with the heading
- Add test for formatting age as expected

## Testing
- View the Directory Service component and test the formatting of the summary list and the alignment looks correct
- Check that the ages are showing as expected
- Use `yalc publish` to publish it locally and then on the frontend `yalc add northants-design-system && yarn install && yarn dev` then view a listing on the frontend and test the ages appear as expected and the formatting is correct with real data.

### DIFF
--- a/src/library/directory/DirectoryService/DirectoryServiceTransform.ts
+++ b/src/library/directory/DirectoryService/DirectoryServiceTransform.ts
@@ -1,5 +1,6 @@
 import { SummaryRowProps } from '../../components/SummaryList/SummaryList.types';
 import {
+  EligibilitiesProps,
   LanguagesProps,
   LocationProps,
   ServiceAreaProps,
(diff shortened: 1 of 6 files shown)


src/library/directory/DirectoryService/DirectoryServiceTransform.ts
```diff
@@ -163,3 +197,16 @@ export const transformTaxonomies = (service_taxonomys: ServiceTaxonomy[], taxono
 
   return details;
 };
+
+export const transformAge = (age: number, ageInMonths: boolean = false) => {
```
chrisrhymes, line 201: Extracted this into a reusable function so it can be used by both the DirectoryServiceList and the DirectoryService components.

PR #376, opened 2023-09-14
Bin Collection Search Component for Northants

drupler:
Bin Collection Search widget:
- Postcode field
- Search for addresses from the provided Drupal API endpoint
- Render a list of addresses in select
- Selecting an address render list of calendar events by fetching from another endpoint

### DIFF
--- a/src/library/components/BinCollection/BinCollection.mock.ts
+++ b/src/library/components/BinCollection/BinCollection.mock.ts
@@ -1,0 +1,80 @@
+// Mock data for address search based on postcode 'NN16 9XR'
+export const addressMockData = {
+    "NN16 9XR": {
+      recordCount: 2,
+      data: [
+        {
+          UPRN: "100031057713",
+          USRN: "18704881",
+          Address: "1 ELIOT CLOSE, KETTERING, NN16 9XR",
+          HouseName: null,
+          AdditionalHouseName: null,
+          Street: "ELIOT CLOSE",
+          Locality: "",
+          Town: "KETTERING",
+          Postcode: null,
+          Longitude: 0,
+          Latitude: 0,
+          UserDefinedAddress: "",
+          CalculatedAddress: null
+        },
+        {
+          UPRN: "100031057722",
+          USRN: "18704881",
+          Address: "10 ELIOT CLOSE, KETTERING, NN16 9XR",
+          HouseName: null,
+          AdditionalHouseName: null,
+          Street: "ELIOT CLOSE",
+          Locality: "",
+          Town: "KETTERING",
+          Postcode: null,
+          Longitude: 0,
+          Latitude: 0,
+          UserDefinedAddress: "",
+          CalculatedAddress: null
+        }
+      ]
+    }
+  };
+  
+// Mock data for calendar events based on UPRN
+export const calendarEventMockData = {
+    "100031057713": [
+      {
+        id: 36193423,
+        title: "Empty KBC Bin Refuse bin 240l",
+        subject: "This Premises",
+        start: "/Date(1661986799000)/",
+        end: null,
+        link: "",
+        color: "rgb(0,0,0)",
+        textColor: "rgb(255,255,255)",
+        complete: true,
+        url: "#"
+      },
+      {
+        id: 4758088,
+        title: "Empty Bin Recycling 240l",
+        subject: "This Premises",
+        start: "/Date(1564613999000)/",
+        end: null,
+        link: "",
+        color: "rgb(192,192,192)",
+        textColor: "rgb(0,0,0)",
+        complete: true,
+        url: "#"
+      },
+      {
+        id: 4735057,
+        title: "Empty Bin Recycling Paper Box",
+        subject: "This Premises",
+        start: "/Date(1564613999000)/",
+        end: null,
+        link: "",
+        color: "rgb(192,192,192)",
+        textColor: "rgb(0,0,0)",
+        complete: true,
+        url: "#"
+      }
+    ]
+  };
--- a/src/library/components/BinCollection/BinCollection.stories.tsx
+++ b/src/library/components/BinCollection/BinCollection.stories.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import BinCollection from './BinCollection';
+import { BinCollectionProps } from './BinCollection.types';
+import { Story } from '@storybook/react/types-6-0';
+import MaxWidthContainer from '../../structure/MaxWidthContainer/MaxWidthContainer';
+import { SBPadding } from '../../../../.storybook/SBPadding';
+
+export default {
+  title: 'Library/Components/Bin Collection',
+  component: BinCollection,
+  parameters: {
+    status: {
+      type: 'stable',
+    },
+  },
+};
+
+const Template: Story<BinCollectionProps> = (args) => (
+  <SBPadding>
+    <MaxWidthContainer>
+      <BinCollection {...args} />
+    </MaxWidthContainer>
+  </SBPadding>
+);
+
+export const TemplateBind = Template.bind({});
+TemplateBind.args = {
+  title: 'Enter your postcode for bin collection details',
+  formError: false,
+};
--- a/src/library/components/BinCollection/BinCollection.styles.ts
+++ b/src/library/components/BinCollection/BinCollection.styles.ts
@@ -1,0 +1,76 @@
+import styled from 'styled-components';
+
+interface FormContainerProps {
+  isLoading: boolean;
+}
+
+export const Container = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+`;
+
+export const Label = styled.label`
+  display: block;
+  margin-bottom: 15px;
+`;
+
+export const BinCollectionButtonStyles = styled.div`
+    button {
+      margin: 10px 10px 22px 10px;
+    }
+`;
+
+export const FormContainer = styled.div<FormContainerProps>`
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  ${(props) => 
+      props.isLoading && `
+          opacity: 0.5; 
+          pointer-events: none;
+      `
+  }
+`;
+
+export const DropDownSelectContainer = styled.div`
+  padding-bottom: 20px;
+`;
+
+export const FormInnerContainer = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: flex-start;
+`;
+
+export const FormLabel = styled.label`
+  display: flex;
+  align-items: center;
+  margin-right: 10px;
+`;
+
+export const UPRNPageTitleRow = styled.div`
+  justify-content: center;
+  margin-bottom: 1em;
+  margin-top: 2em;
+  width: 100%;
+`;
+
+export const UPRNPageUPRNRow = styled.div`
+  display: flex;
+  justify-content: left;
+  margin-bottom: .5em;
+  width: 100%;
+`;
+
+export const UPRNPageSectionTitle = styled.div`
+  font-size: 2em;
+  font-weight: 700;
+`;
+
+export const UPRNPageUPRNTitle = styled.div`
+  font-size: 1.5em;
+`;
+
+export const HomePageResetButton = styled.input``;
--- a/src/library/components/BinCollection/BinCollection.tsx
+++ b/src/library/components/BinCollection/BinCollection.tsx
@@ -1,0 +1,187 @@
+import React, { useState } from 'react';
+import { useContext } from 'react';
+import { ThemeContext } from 'styled-components';
+import axios from 'axios';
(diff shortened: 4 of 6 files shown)


src/library/components/BinCollection/BinCollection.mock.ts
```diff
@@ -0,0 +1,80 @@
+// Mock data for address search based on postcode 'NN16 9XR'
```
DamienBurkeWN, line 1: I feel this file would be better named .testdata.ts, which would make its purpose clear without opening it

src/library/components/BinCollection/BinCollection.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const Template: Story<BinCollectionProps> = (args) => (
+  <SBPadding>
+    <MaxWidthContainer>
+      <BinCollection {...args} />
+    </MaxWidthContainer>
+  </SBPadding>
+);
+
+export const TemplateBind = Template.bind({});
```
DamienBurkeWN, line 26: This needs to be better named, e.g. ExampleBinCollectionSearch

src/library/components/BinCollection/BinCollection.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { ThemeContext } from 'styled-components';
+import axios from 'axios';
+import { BinCollectionProps, AddressOption } from './BinCollection.types';
+import * as Styles from './BinCollection.styles';
+import FormWithLine from '../FormWithLine/FormWithLine';
+import Input from '../Input/Input';
+import FormButton from '../FormButton/FormButton';
+import DropDownSelect from '../DropDownSelect/DropDownSelect';
+
+const BinCollection: React.FunctionComponent<BinCollectionProps> = ({
```
DamienBurkeWN, line 12: Missing comment explaining what the component does

src/library/components/BinCollection/BinCollection.tsx
```diff
@@ -0,0 +1,187 @@
...
+  formError = false,
+  baseUrl = process.env.DRUPAL_APP_API_BASE_URL,
+}) => {
+  const BinCollectionApiUrl = `${baseUrl}bin-collection-search/premiselist/`;
```
DamienBurkeWN, line 16: Why's there a hard-coded postfix to each of these URLs? Shouldn't the URL be a complete bit of configuration?

src/library/components/BinCollection/BinCollection.tsx
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  const [calendarEvents, setCalendarEvents] = useState({});
+
+  const handleAddressChange = (e) => {
+    const selectedUPRN = e.target.value;
+    
+    if (selectedUPRN) {
+      const startDate = getFormattedDate();
+      const endDate = getFormattedDate(42);
```
DamienBurkeWN, line 89: Why 42?

src/library/components/BinCollection/BinCollection.tsx
```diff
@@ -0,0 +1,187 @@
...
+    setAddressOptions([]);
+    setCalendarEvents({});
+    setError(false);
+    setErrorText('');
+  };
+
+  const formatTitle = (apiTitle: string) => {
+    if (apiTitle.includes('Bin Refuse bin')) return 'Refuse bin';
+    if (apiTitle.includes('Bin Recycling')) return 'Recycling bin';
+    if (apiTitle.includes('Bin Recycling Paper Box')) return 'Recycling Paper Box bin';
```
DamienBurkeWN, line 113: This will never fire, as it would be satisfied by the line above

src/library/components/BinCollection/BinCollection.tsx
```diff
@@ -0,0 +1,187 @@
...
+      <FormWithLine
+        onSubmit={handleSubmit}
+        isError={isError}
+        lineColour={themeContext.theme_vars.colours.grey_dark}
+        hideLine
+      >
+        <Styles.FormContainer isLoading={isLoading}>
+          <Styles.FormInnerContainer> 
+            <Styles.Label htmlFor="postcode">
+              Bin Collection Cheker
```
DamienBurkeWN, line 129: Typo - Checker

src/library/components/BinCollection/BinCollection.tsx
```diff
@@ -0,0 +1,187 @@
...
+                type="text"
+                placeholder="Enter a postcode"
+                name="postcode"
+                errorText={errorText}
+                isErrored={isError}
+                maxLength={10}
+              />
+            </Styles.Label>
+            <Styles.BinCollectionButtonStyles>
+              <FormButton size='large' type="submit" aria-label="Submit" text="Find" />
```
DamienBurkeWN, line 140: Minor code style thing - decide on ' or " for attributes not a mixture?

src/library/components/BinCollection/BinCollection.mock.ts
```diff
@@ -0,0 +1,80 @@
...
+        start: "/Date(1564613999000)/",
+        end: null,
+        link: "",
+        color: "rgb(192,192,192)",
+        textColor: "rgb(0,0,0)",
+        complete: true,
+        url: "#"
+      }
+    ]
+  };
```
DamienBurkeWN, line 80: Missing final blank line

DamienBurkeWN (commented):
Unfortuantely am unable to test as no test instructions are provided (e.g. a functioning URL to plug in so that there is an API response) - see descriptions on previous PRs for examples of how best to deliver a more easily testable PR.

As a general note there's a lack of explanatory comments in the code.

As West will need a similar component in the near future it would be worth cooperating on this one more closely, so that we need only change the API URLs for instance, rather than needing a different component.

Will leave Chris to add his review too.

src/library/components/BinCollection/BinCollection.tsx
chrisrhymes: We have found it easier to use [Prettier](https://prettier.io/docs/en/editors.html) in VSCode to auto format items like this. You can enable it to run on save and it formats your code for consistency every time you save.

src/library/components/BinCollection/BinCollection.tsx
```diff
@@ -0,0 +1,187 @@
...
+            <React.Fragment key={title}>
+              <Styles.UPRNPageTitleRow>
+                <Styles.UPRNPageSectionTitle>{title}</Styles.UPRNPageSectionTitle>
+              </Styles.UPRNPageTitleRow>
+              {(dates as Date[])
+                .sort((a, b) => new Date(a).getTime() - new Date(b).getTime())
+                .map(date => (
+                  <Styles.UPRNPageUPRNRow key={`${title}-${date.toString()}`}>
+                    <Styles.UPRNPageUPRNTitle>
+                      {date.toLocaleDateString('en-US', { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' })}
```
chrisrhymes, line 170: Shouldn't we be using `en-GB` format for date formats?

chrisrhymes (commented):
Some additional thoughts on this PR:
- As Damien stated, we cannot test this component as it is. 
- How will the content editor add this component to their page in Drupal? I think this needs to be in a slice rather than a component, or at lease have a slice that loads in this component. This then requires further work in Drupal to create a new custom paragraph that can be selected in the text editor and the frontend updated to pull in the slice. 
- We don't currently call the Drupal backend directly from the client side, we only call external APIs. I think the Drupal backend should only return data via GraphQL for consistency, otherwise we end up with multiple endpoints to manage in Drupal. This means we can't then use axios to fetch the data from Drupal as it will need to be crafted as a GraphQL request.

drupler:
@DamienBurkeWN This component would work in conjunction with Drupal application endpoints, which basically connect to the Bartec system and return a list of addresses (provided a postcode) and a list of calendar events given a UPRN number. 

For testing, you can pull the branch from this PR https://github.com/FutureNorthants/northants-website/pull/843 and use your localhost path in the env variable named 'DRUPAL_APP_API_BASE_URL'. In the storybook component, providing a correct postcode and selecting should result in something like:

Screenshot from 2023-09-18 09-43-13

Let me know if you face any trouble in this. Thanks

src/library/components/BinCollection/BinCollection.mock.ts
drupler: @DamienBurkeWN Done that, thanks

src/library/components/BinCollection/BinCollection.stories.tsx
drupler: @DamienBurkeWN Done that, thanks

src/library/components/BinCollection/BinCollection.tsx
drupler: @DamienBurkeWN Done that, thanks

src/library/components/BinCollection/BinCollection.tsx
drupler: @DamienBurkeWN Two different routes are created in the Drupal application, which connects to Bartec to fetch the list of addresses and calendar events used here. PR for that is here -- https://github.com/FutureNorthants/northants-website/pull/843. Please review and suggest if you think we should adopt something else. Thanks

src/library/components/BinCollection/BinCollection.tsx
drupler: @DamienBurkeWN Bartec API doesn't work on date filtering yet, so according to the suggestion of JK, it's being restricted to the next 42 days from here when rendering in the frontend.

src/library/components/BinCollection/BinCollection.styles.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export const Container = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+`;
+
+export const Label = styled.label`
+  display: block;
+  margin-bottom: 15px;
```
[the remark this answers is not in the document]
drupler: @chrisrhymes Changed to that, thanks

src/library/components/BinCollection/BinCollection.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  const resetForm = () => {
+    setAddressOptions([]);
+    setCalendarEvents({});
+    setError(false);
+    setErrorText('');
+  };
+
+  const formatTitle = (apiTitle: string) => {
+    if (apiTitle.includes('Bin Refuse bin')) return 'Refuse bin';
```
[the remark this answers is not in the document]
drupler: @chrisrhymes Yes, it does based on UPRN. However, I have made it more concise which makes it easier to maintain. Thanks

src/library/components/BinCollection/BinCollection.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { BinCollectionProps, AddressOption } from './BinCollection.types';
+import * as Styles from './BinCollection.styles';
+import FormWithLine from '../FormWithLine/FormWithLine';
+import Input from '../Input/Input';
+import FormButton from '../FormButton/FormButton';
+import DropDownSelect from '../DropDownSelect/DropDownSelect';
+
+const BinCollection: React.FunctionComponent<BinCollectionProps> = ({
+  formError = false,
+  baseUrl = process.env.DRUPAL_APP_API_BASE_URL,
```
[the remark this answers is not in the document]
drupler: @chrisrhymes Changes done like you suggested. Thanks

src/library/components/BinCollection/BinCollection.mock.ts
drupler: @DamienBurkeWN Fixed that, thanks

src/library/components/BinCollection/BinCollection.tsx
drupler: @DamienBurkeWN Correction done. Thanks

src/library/components/BinCollection/BinCollection.tsx
drupler: @DamienBurkeWN Correction done. Thanks

src/library/components/BinCollection/BinCollection.tsx
drupler: @DamienBurkeWN Nice catch. Changing it to 'Recycling Paper Box' now. Thanks

chrisrhymes:
I don't see this working with a boolean field on a service page as it may need to have other content above or below it, rather than a fixed point on a page. Also, it will probably only ever be on one page so making it a page setting doesn't really make sense to me. 

The other large issue with this is that it assumes Bartec is the only waste provider. There are at least three different waste providers in West Northants so this will not return any data for many residents as they don't use Bartec. I am unsure how many providers North Northants has, but if there are areas that use another provider then they won't get any data either. 

It is not a very flexible approach either as what happens if the provider changes their response format or we get another provider added in future? The frontend should not really care who the provider is, it should just display the information provided. 

I think this needs to be reconsidered as part of a larger piece of work for all waste providers.

PR #377, opened 2023-10-05
Update back link to allow using browser history

chrisrhymes:
Users can get stuck in a loop if there are two pages in a row with a back link on as it adds to the browser history (as a link) instead of removing an item from the browser history. 

This change allows you to set `useHistory` to go back using the browser history instead. 

When useHistory is true it will be a button to use the onClick event. When it is a link it will be a standard anchor tag. 

## Testing
- Checkout this branch and run `npm run dev` to test the two stories
- The useHistory should navigate you back to the previous page you were on and the other should be a standard link
- Run `npm run test` to manually run the test suite

### DIFF
--- a/src/library/structure/BackLink/BackLink.stories.tsx
+++ b/src/library/structure/BackLink/BackLink.stories.tsx
@@ -1,24 +1,31 @@
-
-import React from "react";
+import React from 'react';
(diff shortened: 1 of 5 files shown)


DamienBurkeWN (approved):
👍

PR #382, opened 2023-11-03
Add error states for directory list components

chrisrhymes:
Previously the error state was ignored and just displayed no results. This adds an error message to the DirectoryDocumentList and the DirectoryServiceList components when the isError prop is true. 

## Testing
- View the stories for DirectoryDocumentList and DirectoryServiceList and there is now a story for error state
- Manually run tests with `npm run test`

## Frontend testing
- Run `yalc publish` and then go to the frontend and run `yalc add northants-design-system && yarn install && yarn dev`
- In the frontend, update src/DirectoryPages/DirectoryPage/index.ts and add 'error' in line 182 `const { loading, error, data } = useQuery(ServicesQuery, `
- Pass in isError prop to the DirectoryServiceList component `isError={!!error}` (this is because error is not true when it exists, it's `ApolloError` type). 
- Stop your local container for open referral
- Visit both the local offer and fis directory search pages and it should now display the error message. 
- Start the open referral container again then refresh the directory search pages. Error message should be gone and the results displaying once more. Ensure loading state still shows as expected.

### DIFF
--- a/src/library/directory/DirectoryDocumentList/DirectoryDocumentList.stories.tsx
+++ b/src/library/directory/DirectoryDocumentList/DirectoryDocumentList.stories.tsx
@@ -42,3 +42,35 @@ ExampleDirectoryDocumentList.args = {
   setCategories: () => {},
   isLoading: false,
 };
+
+export const ExampleDirectoryDocumentListNoResults = Template.bind({});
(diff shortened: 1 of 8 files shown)


DamienBurkeWN (approved):
👍

PR #383, opened 2023-11-21
Add button to accept cookies from the static map

chrisrhymes:
Add a button that allows users to accept cookies on the static map, so that users can choose to interact with a map instead of seeing the static map only. 

- Updated cookie banner to export the Accept and Reject cookies methods
- Update static map to use the accept cookies method

## Testing
- Run `npm run dev` and then clear your cookies for localhost
- Visit the static map and press `accept cookies`. The page should reload.
- Check your cookies and there should now be a cookie with the name `fn-cookie` with an expiry date set to a years time. 

## Frontend testing
- Run `yalc publish` and then in the frontend run `yalc add northants-design-system && yarn install && yarn dev`
- Visit the directory page and clear your cookies
- The static map should show when you press Show Map. 
- Click the `accept cookies` button in the static map message.
- The page should reload and the cookies should now be accepted and the interactive map should now show when you click Show Map.

### DIFF
--- a/src/library/components/StaticMap/StaticMap.styles.js
+++ b/src/library/components/StaticMap/StaticMap.styles.js
@@ -12,3 +12,26 @@ export const Message = styled.div`
   background-color: ${(props) => props.theme.theme_vars.colours.focus};
   padding: ${(props) => props.theme.theme_vars.spacingSizes.medium};
 `;
+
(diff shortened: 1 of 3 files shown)


DamienBurkeWN (approved):
👍

PR #386, opened 2023-11-27
Update help text on directory postcode search

chrisrhymes:
Make the help text for the postcode search more informative to explain to users that it filters results to within 2 miles of the postcode that has been entered. 

Part of an initial fix for FIS-53 (which requires further development). 

## Testing
- Checkout this branch and run `npm run dev` and view the Directory Service List component
- Check the updated help text for the postcode search


DamienBurkeWN (approved):
💪 👍

PR #387, opened 2023-11-28
Format directory age in years an months when relevant

chrisrhymes:
- Update the transformAge function to show the months when it is not an exact age in years. 
- Added tests

## Testing
- Checkout this branch and run `npm run dev` and view the Directory Service List to test how the various ages appear.
- Under 36 months should still show as month
- Over 36 months but not an exact year should show the years and months. If it is 1 month it should show '1 month', but if it is more it should be plural, such as '2 months'
- Over 36 months and an exact year should show the years but not months
- Run the tests manually with `npm run test`


DamienBurkeWN (approved):
👍

PR #388, opened 2023-11-28
Update/node 18 storybook 7

chrisrhymes:
- Update to storybook v7 (requires node > 16 so specified 18)
- Update testing library and jest
- Upgrade typescript version
- Update to styled-components v6
- Migrate to storybook v7 following [migration guide](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-65x-to-700)
- Update .storybook/main.js and preview.js to typescript files and follow new export structure
- Update Story type to StoryFn in stories
- Remove GitHub flavoured markdown addon

## Testing
- **Use node v18!** I use [nvm](https://github.com/nvm-sh/nvm) to install multiple versions of node on my dev machine. 
- Checkout this branch and run `npm install` to update dependencies
- Run `npm run test` to manually run test suite. There should be no warnings about [Unknown Props](https://legacy.reactjs.org/warnings/unknown-prop.html) (but not sure every component has a test yet to ensure this!)
- Run `npm run dev` to launch storybook for manual testing. There should be no warnings in the console when the command is run.


DamienBurkeWN (approved):
If I wasn't cross-eyed before this PR I would have been by now.

PR #391, opened 2023-12-15
Replace react-progressive-lazy-image with next/image

chrisrhymes:
- Built upon #388 
- Add @storybook/nextjs and next to dev dependencies
- Make use of next/image and replace the abandonend react-progressive-lazy-image to display images, use the small image as a placeholder and lazy load the large image when scrolling into view

**Has issues running with the frontend when using `yarn dev`!**

## Testing
- Checkout this branch and run `npm install` then `npm run dev` to manually test updated components.

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -7,8 +7,12 @@ const config: StorybookConfig = {
   },
 
   framework: {
-    name: '@storybook/react-webpack5',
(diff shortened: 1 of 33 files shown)


chrisrhymes:
I don't think this is the way forward so closing this branch.

PR #392, opened 2023-12-19
Fix/breadcrumbs list when multiple

chrisrhymes:
Resolves https://github.com/FutureNorthants/northants-website/issues/874

Monsido is warning that you should not have a single link as a list, so this checks if there is only one and makes it a div and span instead of ol and li. 

## Testing
- View the Structure -> Breadcrumbs stories and inspect the code. 
- When there is only one breadcrumb it should not be a list (div and span).
- When there are multiple breadcrumbs it should be a list.


DamienBurkeWN (approved):
👍

PR #394, opened 2024-01-04
Increase link sizes for WCAG 2.1 AAA 2.5.5

chrisrhymes:
This is to help meet the 'Aim for large interactive controls' (WCAG 2.1 AAA 2.5.5) for the elements in the header and footer. This was picked up on the [Silktide report](https://app.index.silktide.com/silktide-index/everything/inspector?checkId=ensure-reasonable-target-size&device=mobile&pointId=9cec3a9e0515d6eff004ee2d1f0f1955&scId=11872587389b460442eaa4957668e1b6&websiteId=3742). The minimum size to aim for to reach AAA is 44px by 44px. I have tried to add padding to the links and then removed the same amount from the containers so they still look the same to the end user. 

The footer links are now stacked on mobile (block instead of inline-block) so that they are easier to read and click. 

This branch is based on the [storybook 7 branch](https://github.com/FutureNorthants/northants-design-system/pull/388) so needs node v18. 

## Testing
- View the Header, Footer and Breadcrumbs components and test in various screen sizes to test the responsiveness
- Ensure that the links are now all a minimum of 44px x 44px.

### DIFF
--- a/src/library/structure/Breadcrumbs/Breadcrumbs.styles.js
+++ b/src/library/structure/Breadcrumbs/Breadcrumbs.styles.js
@@ -1,7 +1,7 @@
 import styled from 'styled-components';
 
 export const Container = styled.div`
-  padding-top: 20px;
+  padding-top: 15px;
   border-bottom: 1px solid ${(props) => props.theme.theme_vars.colours.grey}80;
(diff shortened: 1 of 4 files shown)


DamienBurkeWN (approved):
Works, big search box is a bit ugly, but if that's AAA...

PR #397, opened 2024-01-22
Fix type declarations missing in rollup build

chrisrhymes:
After publishing v1 it seems the type declarations are missing. I think this was working when I tested before as I had a previous build with the types included and they were not removed by the build. When I deleted the build folder and published, the type declarations are missing. 

This reverts changes to the tsconfig that were made due to issues I was having with jest not finding the types. This doesn't seem to be an issue anymore after removing the 'src/**/*.test.tsx' from the exclude array. This means other changes can be reverted. 

## Testing
- Checkout this branch and run `npm install` (using node 18)
- Delete the `build` folder so it creates it all from scratch
- Run `yalc publish`
- Use [this branch](https://github.com/FutureNorthants/northants-website/pull/872) in the frontend and run `yalc add northants-design-system` then `npm install` and `npm run build`. 
- The frontend build should now work without type declarations missing from the build

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
+    "rootDir": "src",
     "declaration": true,
-    "declarationDir": "build",
+    "declarationDir": "./build",
     "module": "esnext",
     "target": "es5",
     "lib": ["es6", "dom", "es2016", "es2017"],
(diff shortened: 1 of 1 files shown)


chrisrhymes:
See changes that were [made here](https://github.com/FutureNorthants/northants-design-system/pull/388/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0L3) that caused the issue to occur.

DamienBurkeWN (approved):
👍

PR #401, opened 2024-02-08
Fix search bar alignment with error message

chrisrhymes:
When the search bar was within PageMain, the error message had more margin. This overrides the margin so it is fixed. The margin also needed to be changed based on the screen width as the error text changes size too. 

## Testing
- Test out the stories for the search bar. Some use page main and others don't. 
- Test different screen sizes to ensure the button aligns correctly when the error message is shown (search without a term).


DamienBurkeWN (approved):
👍

PR #406, opened 2024-02-09
Fix link active styles top

### LINKED ISSUE #405: Cannot click the top of the buttons as they move using CSS transform
When you try and click the very top of a button with linkStylesActive in the theme_generator, it moves the button down using transform: translate. This means you are no longer pressing the button so it doesn't trigger the click event properly. 

The Gov UK design system does something similar, but uses `top: 2px` instead of a transform. Test this method out to see if it works as expected and doesn't cause unintended consequences.

chrisrhymes:
Resolves #405 

The button was previously being moved using transform: translate so it was no longer under the cursor when clicking the top of the button. This now uses top to move the button. 

## Testing
- Using master branch, run `npm run dev` and view the Directory Service List 
- Click the top of the 'Add to shortlist' or 'Remove from shortlist' button and notice the button does not toggle after click
- Repeat the test and the button should now toggle state.

### DIFF
--- a/src/themes/theme_generator.tsx
+++ b/src/themes/theme_generator.tsx
@@ -69,7 +69,7 @@ const generate_theme = (theme_vars: ThemeVars) => {
       background: ${theme_vars.colours.focus};
       outline: none;
       text-decoration: none !important;
-      transform: translateY(2px);
+      top: 2px;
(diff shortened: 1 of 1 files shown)


DamienBurkeWN (approved):
👍

PR #414, opened 2024-02-14
Replace LazyImage with ResponsiveImage

### LINKED ISSUE #390: Replace react-lazy-progressive-image
The react-lazy-progressive-image repo has been archived. We need to replace it with an alternative. 

Investigate [@storybook/nextjs](https://www.npmjs.com/package/@storybook/nextjs) for use with Storybook v7 as it supports Next Image which offers a lot of useful features.

chrisrhymes:
Resolves #390

The react-lazy-progressive-image package is deprecated. This PR replaces it with a custom component called ResponsiveImage. 

- It uses a standard image tag and makes use of the `loading="lazy"` attribute to make the images load when scrolled into view. More information [here](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading).
- It uses `srcset` to display smaller images on smaller screen resolutions where possible. 
- It allows you to pass in the image ratio, but also allows you to set whether the image should cover the container or be contained within the container. The inspiration came from [here](https://www.smashingmagazine.com/2021/10/object-fit-background-size-css/)
- It has a ratio called 'auto' that allows the image container to fit the parent container. When using this you need to ensure the parent container has `position: relative;`. This is used on the search box, hero and home hero.
- When it is used instead of a background image, the other containers have been updated to use a higher z-index to make the elements sit above the image. I considered using `background-image: image-set()` instead ([more info here](https://developer.mozilla.org/en-US/docs/Web/CSS/image/image-set)), but this is for screen quality, rather than screen resolutions. 

## Testing
- Run `npm run dev` to manually test the affected components.
- View the Example Content page. Inspect the network tab and scroll down. You should see the images appear when they scroll into view. 
- Run `npm run test` to run the test suite.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,6 @@
         "core-js": "^3.20.2",
         "downshift": "^6.1.7",
         "html-react-parser": "^3",
-        "react-lazy-progressive-image": "^1.5.5",
         "react-qr-code": "^2.0.11",
         "sanitize-html": "^2.6.1",
         "stylis": "^4.3.0"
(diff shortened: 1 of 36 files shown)


DamienBurkeWN (approved):
Niiiiiiice. Much better 👍

PR #425, opened 2024-03-05
Add dropdown for postcode radius search on directory service list

chrisrhymes:
- Add distance dropdown to allow users to change the default radius of 2 miles. 
- Simplify the layout removing hint text, pagination from the top, clear filters button from the top of the filters (as now have Clear search button)
- Update DropDownSelect component to allow for bold label text and optional hint text.

## Testing
- View the Directory Service List component to view the visual changes.


src/library/directory/DirectoryServiceList/DirectoryServiceList.tsx
```diff
@@ -219,15 +221,31 @@ const DirectoryServiceList: React.FunctionComponent<DirectoryServiceListProps> =
...
+                    </Styles.PostcodeInner>
+                    <DropDownSelect
+                      id="proximity"
+                      label="Distance"
+                      options={[
+                        { title: '', value: '2' },
+                        { title: '2 miles', value: '2' },
+                        { title: '5 miles', value: '5' },
+                        { title: '10 miles', value: '10' },
+                        { title: '20+ miles', value: '50' },
```
DamienBurkeWN, line 243: Description and limit don't match - should this be a limit of 20 or a limit of 50?

DamienBurkeWN (changes requested):
Dropdown isn't defaulting to 2 miles, it's empty.
Dropdown is appearing grey - I'd assumed disabled - but it isn't.
Should dropdown be useable if postcode isn't entered?

src/library/directory/DirectoryServiceList/DirectoryServiceList.tsx
chrisrhymes: The radius has to be an integer, so setting to 50 to include most of West Northamptonshire. 

The new meilisearch will use the [_geoRadius](https://www.meilisearch.com/docs/learn/fine_tuning_results/geosearch#_georadius) method so needs a distance value (the backend will turn the miles into metres.

DamienBurkeWN: So shouldn't the title be '50 miles'? As 51 is '20+' but wouldn't be included.

chrisrhymes:
More research and feedback needed for this feature so converting to a draft pull request.

PR #429, opened 2024-03-06
Hide directions button when no lat lon

chrisrhymes:
Resolves https://github.com/FutureNorthants/northants-website/issues/913 

Hide the 'Get directions' button on the DirectoryService when there is no latitude or longitude. 

## Testing
- View the Example Directory Service Visitable No Lat Lon and ensure the map does not show and there is no 'Get directions' button.

### DIFF
--- a/src/library/directory/DirectoryService/DirectoryService.tsx
+++ b/src/library/directory/DirectoryService/DirectoryService.tsx
@@ -122,10 +122,12 @@ const DirectoryService: React.FunctionComponent<DirectoryServiceProps> = ({
                                 .join(' <br />'),
                             }}
                           />
-                          <Button
-                            url={`https://google.com/maps/dir//${location.latitude},${location.longitude}`}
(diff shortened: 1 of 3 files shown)


src/library/directory/DirectoryService/DirectoryService.tsx
```diff
@@ -122,10 +122,12 @@ const DirectoryService: React.FunctionComponent<DirectoryServiceProps> = ({
                                 .join(' <br />'),
                             }}
                           />
-                          <Button
-                            url={`https://google.com/maps/dir//${location.latitude},${location.longitude}`}
-                            text="Get directions"
-                          />
+                          {location.latitude && location.longitude && (
```
DamienBurkeWN, line 125: Well we better just hope we have no Northamptonshire services located on Null Island in the Atlantic, pal.

DamienBurkeWN (approved):
👍

PR #450, opened 2024-04-03
Update to storybook v8 and replace babel with swc

chrisrhymes:
- Update to storybook v8 using the [migration guide](https://storybook.js.org/docs/migration-guide)
- Replace babel with [SWC](https://storybook.js.org/docs/configure/compilers#swc) for faster builds and [tests](https://swc.rs/docs/usage/jest)
- Update other dependencies to latest versions
- Remove unused dependencies
- Update GitHub workflows to use node 18

## Testing
- Manually test with `npm run dev`
- Run test scripts with `npm run test`


src/library/structure/AccessibleLinks/AccessibleLinks.tsx
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import { AccessibleLinkProps } from "./AccessibleLinks.types";
 import * as Styles from "./AccessibleLinks.styles";
 
-const AccessibleLink: React.FC<AccessibleLinkProps> = ({
+const AccessibleLinks: React.FC<AccessibleLinkProps> = ({
```
chrisrhymes, line 5: This is imported as AccessibileLinks in src/library/structure/PageStructures.ts and was causing the build to fail. Updating the component name to the file name fixed it.

.storybook/main.ts
```diff
@@ -30,46 +29,6 @@ const config: StorybookConfig = {
     NEXT_PUBLIC_BIN_FINDER_API_URL: 'https://api.westnorthants.digital/openapi/v1/unified-waste-collections/',
   }),
 
-  webpackFinal: async (config) => {
```
chrisrhymes, line 33: We don't use sass or scss as we are using styled components. Also removed babel config now we are using SWC instead.

DamienBurkeWN (approved):
👍

PR #451, opened 2024-04-03
Force page break on print after 2 items

chrisrhymes:
Prevent services from going across multiple pages when printing, forcing the page break after every other item in the shortlist. 

## Testing
- Checkout this branch and run `npm run dev`.
- View the Directory Service List and add all services to the shortlist.
- View the Directory Shortlist and press print. Services should not be split across pages.


DamienBurkeWN (approved):
👍

PR #465, opened 2024-04-23
Store UPRN and address as cookie when cookies accepted

### LINKED ISSUE #464: Remember your address in the bin finder
If opted in to cookies, then store the last UPRN found so it can fetch the data automatically when the page is refreshed. This will save time and prevent the need for address lookup leading to quicker results.

chrisrhymes:
Resolves #464 

Once you have found an address and it's UPRN (and cookies were accepted), it will store these in a cookie that can be retrieved next time the page is loaded. This prevents the need to get the address each time and save users time. 

## Testing
- Clear the cookies for the site to start fresh.
- Visit the Page Examples -> Bin Collection Day -> Bin Finder Example. You should see the cookies banner at the top of the page, indicating cookies have not been accepted. 
- Do a search for a postcode and address as you would normally and get the bin collection details as you would normally. 
- Refresh the page and the form should reset to asking for the postcode once more and not show your collection details.
- Now click Accept Cookies and the page should reload.
- Repeat the search as before and get the bin collection details.
- Refresh the page and the bin collection details should now load in automatically.

### DIFF
--- a/src/library/pages/BinFinderExample/BinFinderPageExample.tsx
+++ b/src/library/pages/BinFinderExample/BinFinderPageExample.tsx
@@ -11,6 +11,18 @@ export interface BinFinderPageExampleProps {
 export const BinFinderPageExample: React.FunctionComponent<BinFinderPageExampleProps> = ({ title, contactInfo }) => {
   return (
     <>
+      <PageStructures.CookieBanner
+        title="Tell us whether you accept cookies"
+        paragraph={
+          <p>
+            We use <a href="/your-council/cookies">cookies to collect information</a> about how you use this website. We
+            use this information to make it work as well as possible and help make improvements.
+          </p>
+        }
+        acceptButtonText="Accept all cookies"
+        rejectButtonText="Reject all cookies"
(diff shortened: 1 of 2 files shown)


src/library/pages/BinFinderExample/BinFinderPageExample.tsx
```diff
@@ -11,6 +11,18 @@ export interface BinFinderPageExampleProps {
 export const BinFinderPageExample: React.FunctionComponent<BinFinderPageExampleProps> = ({ title, contactInfo }) => {
   return (
     <>
+      <PageStructures.CookieBanner
+        title="Tell us whether you accept cookies"
+        paragraph={
+          <p>
+            We use <a href="/your-council/cookies">cookies to collect information</a> about how you use this website. We
```
DamienBurkeWN, line 18: We could probably reword this to be a bit friendlier, mention things like bin finder remembering you perhaps?

DamienBurkeWN (approved):
ABSOLUTELY GODDAMN FANTASTIC MAN 👍

PR #471, opened 2024-04-29
Update postcode search in bin finder slice to begins with

### LINKED ISSUE #468: Bin finder - update search to be begins with rather than contains
Testing feedback:

> E.g. search for NN1 5PH and 'Room 12' - nope. '12' - nope. '27' - bingo, includes the Room 12 we wanted? I would expect it show all the addresses with Room 12 if I search for Room 12?

To help reduce options returned from the bin finder address lookup, and therefore the number of pages returned, it would be better to have a begins with search rather than a contains search. This should help users find their address easier. 

For example, searching for 1 would return many matches when the postcode has a 1 in, such as nn1 or nn12. A begins with search should return house number 1, 11, 12, etc. 

An option to increase the amount of results per page should also allow the user to select from a dropdown when there are slightly more values, such as 30 rather than 25.

chrisrhymes:
- Resolves #468 
- Also add autocomplete to the input fields to allow easier autocompletion by the browser. 

## Testing
- Checkout this branch and update `NEXT_PUBLIC_POSTCODE_SEARCH_API_URL` in .storybook/main.ts to dev endpoint (ask Chris)
- Run `npm run dev`
- View the Slices -> Example Bin Finder
- Search for postcode `NN1 5PH` and press Find
- In the address search, enter `Room 12` and press Find. It should now find matches beginning with Room 12 and allow you to select a result from the dropdown and press Submit.

### DIFF
--- a/src/library/slices/BinFinder/BinFinder.tsx
+++ b/src/library/slices/BinFinder/BinFinder.tsx
@@ -102,7 +102,7 @@ const BinFinder: React.FunctionComponent<BinFinderProps> = ({ title, contactInfo
     const formattedHouseNumber = data.houseNumber.trim();
 
     axios
-      .get<PostcodeResultsProps>(`${PostcodeSearchApiUrl}${formattedPostcode}?address=${formattedHouseNumber}`)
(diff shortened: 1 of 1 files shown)


DamienBurkeWN (approved):
👍

PR #476, opened 2024-05-02
Update downshift to v9

### LINKED ISSUE #348: Update downshift to v9
Downshift is an accessible autocomplete, used in the Autocomplete component in the design system. 

Downshift v7 was released in October last year and contains potentially breaking changes. Follow the [migration guide](https://github.com/downshift-js/downshift/blob/master/src/hooks/MIGRATION_V7.md) to update from v6 and test as needed.  

Use `yalc` to publish locally and test with the frontend.

chrisrhymes:
Resolves #348 

Updated downshift to the latest version (9 now rather than 8 as the issue states). I read through the migration guides and they seem to be focused on changes to the useCombobox and useSelect hooks which we are not using. The only change made was to replace undefined with an empty object as typescript was giving an error. 

## Testing

- Checkout this branch and run `npm install` to update dependencies
- Run `npm run dev` and view the Autocomplete stories to manually test the autocomplete works as expected
- Run `npm run test` to run the test suite
- Run `yalc publish` then in the frontend run `yalc add northants-design-system && npm install && npm run dev` to manually test the search bar shows suggestions and searches when an option is selected.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "ISC",
       "dependencies": {
         "core-js": "^3.20.2",
-        "downshift": "^6.1.7",
+        "downshift": "^9.0",
         "html-react-parser": "^5",
         "react-qr-code": "^2.0.11",
         "sanitize-html": "^2.6.1",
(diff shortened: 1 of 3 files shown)


DamienBurkeWN (approved):
👍

PR #477, opened 2024-05-07
Coroners inquest schedule redesigned layout

chrisrhymes:
Resolves https://github.com/FutureNorthants/northants-website/issues/959

Replaced the table layout as it was difficult to use on smaller screen sizes. The schedule is now grouped by day and then displayed in time order, with the data in a column layout, instead of in a table. 

## Testing
- Visit the Slices -> Inquest Schedule slice to manually test the layouts, for when there is data, when there isn't data and when an error occurs. 
- Test layouts on mobile devices to ensure content is still readable on smaller screens. 
- Run test script with `npm run test`

### DIFF
--- a/src/library/slices/InquestSchedule/InquestSchedule.tsx
+++ b/src/library/slices/InquestSchedule/InquestSchedule.tsx
@@ -57,13 +57,14 @@ const InquestSchedule: React.FunctionComponent<InquestScheduleProps> = ({ caseAp
                           <strong>{formatTime(startDateTime)}</strong>
                         </Styles.InquestTime>
                         <Styles.InquestDetails>
-                          {inquest.fullName}.
+                          <strong>Name:</strong> {inquest.fullName}.
                           <br />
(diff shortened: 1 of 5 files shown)


chrisrhymes:
Customer has now approved the updated layout so is ready for code review.

src/library/slices/InquestSchedule/InquestSchedule.tsx
```diff
@@ -1,54 +1,92 @@
...
+                  );
+                })}
+            </Row>
+          </div>
+        );
+      })}
+
+      {caseAppointments.length === 0 && (
+        <div>
+          <p>We can't find any results at the moment.</p>
```
DamienBurkeWN, line 80: I know it was already there but this line seems superfluous, and potentially confusing (if there's no results as there are no inquests next month, the "at the moment" seems to imply if I refresh I might have more luck...?). Anyway not a biggy.

DamienBurkeWN (approved):
👍

PR #479, opened 2024-05-10
Fix May 24 silktide accessibility findings

### LINKED ISSUE #478: Silktide May 24 accessibility findings
- Homepage promoted links insufficient contrast on focus - Ensure controls clearly indicate when they are selected
- HeroBanner text without the box has insufficient contrast with the background. This is probably due to replacing the lazy image component with the custom ResponsiveImage component. Maybe add the background overlay to the hero rather than the responsive image component?

chrisrhymes:
Resolves #478 

- Update the focus styles on PromotedLinks so they have both black and gold for increased contrast. It has also been updated to surround the whole link, rather than just the bottom of the link, to make the focus selection more obvious. 
- Set a background colour to the HeroImage container to ensure there is a colour contrast background for the white text. It looks like the automated scan cannot determine the background colour as it is an image. Any other recommendations for this issue are welcome. 

## Testing
- View the Pages -> Homepage, then click into the search bar. Press tab to move focus to the promoted links and test the contrast when focused.
- View the Pages -> Service Landing Page -> Micro Site Example, then check the image and text still display as expected. Check other Micro Site stories to ensure they still behave as expected. There may be a flash of the dark grey background before the image loads because the images are designed to lazy load.

### DIFF
--- a/src/library/components/PromotedLinks/PromotedLinks.styles.js
+++ b/src/library/components/PromotedLinks/PromotedLinks.styles.js
@@ -35,11 +35,14 @@ export const PromotedLink = styled.a`
   background: ${(props) => props.theme.theme_vars.colours.white};
   background: ${(props) => props.theme.theme_vars.colours.white}F2;
   border-radius: 3px;
-  box-shadow: 0px -4px 0px 0px ${(props) => props.theme.theme_vars.colours.action} inset,
+  box-shadow:
+    0px -4px 0px 0px ${(props) => props.theme.theme_vars.colours.action} inset,
     0px 4px 15px rgba(0, 0, 0, 0.11);
-  -webkit-box-shadow: 0px -4px 0px 0px ${(props) => props.theme.theme_vars.colours.action} inset,
+  -webkit-box-shadow:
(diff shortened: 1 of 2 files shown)


src/library/components/PromotedLinks/PromotedLinks.styles.js
```diff
@@ -56,21 +59,31 @@ export const PromotedLink = styled.a`
 
   &:focus {
     ${(props) => props.theme.linkStylesFocus};
-    box-shadow: 0px -4px 0px 0px ${(props) => props.theme.theme_vars.colours.black} inset,
-      0px 4px 15px rgba(0, 0, 0, 0.11);
-    -webkit-box-shadow: 0px -4px 0px 0px ${(props) => props.theme.theme_vars.colours.black} inset,
-      0px 4px 15px rgba(0, 0, 0, 0.11);
-    -moz-box-shadow: 0px -4px 0px 0px ${(props) => props.theme.theme_vars.colours.black} inset,
-      0px 4px 15px rgba(0, 0, 0, 0.11);
+    box-shadow:
```
chrisrhymes, line 62: This is the change to the focus styles. The rest are prettier changes.

DamienBurkeWN (approved):
👍

PR #488, opened 2024-06-03
Update thank you text and reorder content sections

chrisrhymes:
This resolves [feedback from the issue](https://github.com/FutureNorthants/northants-website/issues/879#issuecomment-2145024497)

- Add paragraph of text to the thank you message
- Reorder the sections so the content goes first 
- Update styles on panel component to fix layout bug when text was added. 

## Testing
- View the Structure -> Rate this page -> Successful rate this page
- Select 'Yes' and press 'Submit' to see the updated thank you text. 
- Refresh the page and this time press 'No'. The 'Content' box should now be before the 'Service' box.

### DIFF
--- a/src/library/structure/RateThisPage/RateThisPage.tsx
+++ b/src/library/structure/RateThisPage/RateThisPage.tsx
@@ -140,7 +140,12 @@ const RateThisPage: React.FunctionComponent<RateThisPageProps> = ({
   return (
     <Styles.FormContainer as="section" data-testid="RateThisPage" aria-label="Rate This Page">
       {isSuccessful ? (
-        <Panel heading="Thank you for your feedback." />
+        <Panel heading="Thank you for your feedback.">
+          <p>
+            If you have left your email address, we will be in touch with you about your feedback in the next 5 working
+            days.
+          </p>
+        </Panel>
       ) : (
         <form onSubmit={executeCaptcha} ref={fullFormRef}>
           <Row>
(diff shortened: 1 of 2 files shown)


src/library/structure/RateThisPage/RateThisPage.tsx
```diff
@@ -140,7 +140,12 @@ const RateThisPage: React.FunctionComponent<RateThisPageProps> = ({
   return (
     <Styles.FormContainer as="section" data-testid="RateThisPage" aria-label="Rate This Page">
       {isSuccessful ? (
-        <Panel heading="Thank you for your feedback." />
+        <Panel heading="Thank you for your feedback.">
+          <p>
+            If you have left your email address, we will be in touch with you about your feedback in the next 5 working
```
DamienBurkeWN, line 145: If I say yes, this page was helpful, then email was never asked for and there's no reason to be in touch...

DamienBurkeWN (changes requested):
Thank you message is inappropriate if you're actually happy with the page

src/library/structure/RateThisPage/RateThisPage.tsx
chrisrhymes: Good spot! I have now wrapped this in an if so it only displays if you select. 'No' to the first question.

DamienBurkeWN (approved):
👍

PR #496, opened 2024-06-20
Add web chat bot form

chrisrhymes:
See this issue for details: https://github.com/FutureNorthants/northants-website/issues/1030

The web chat will only be accessible from the contact page. This means that it needs to be a slice so it can be added as page content in the CMS. 

- [WebChat slice component](https://deploy-preview-496--northants-design-system.netlify.app/?path=/story/library-slices-webchat--example-web-chat)
- [Contact page preview](https://deploy-preview-496--northants-design-system.netlify.app/?path=/story/page-examples-contact-page--example-contact-page)

The frontend will need to implement code to check whether there are any agents available and what channels can be contacted. This is done when the user presses the 'Open WebChat' button. It is currently simulated by using a set timeout to show the loading spinner, but will need to be updated to contact the real API and get the required data. 

## Testing
- View the Slices -> WebChat -> Example Web Chat story
- Update the action prop to the testing endpoint (ask in teams for it) then test completing the form.
- The name, subject and enquiry type should be required fields. It should not submit without these values.
- When you submit it should open a new tab and take you to the Storm web chat with the fields you completed in the url. The enquiry type should use the 'queue' query parameter.


src/library/components/DropDownSelect/DropDownSelect.tsx
```diff
@@ -14,26 +14,46 @@ const DropDownSelect: React.FC<DropDownSelectProps> = ({
...
-        $isErrored={isErrored}
-      >
-        {options.map((option, i) => (
-          <Styles.Option key={i} value={option.value}>
-            {option.title}
-          </Styles.Option>
-        ))}
-      </Styles.Select>
+      {errorText && <Styles.ErrorText id={`${id}Error`}>{errorText}</Styles.ErrorText>}
+      {typeof value !== 'undefined' ? (
```
chrisrhymes, line 27: typeof returns a string, so updated this line to fix the bug where it wouldn't set the default value.

src/library/slices/WebChat/WebChat.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+      errorMessage: 'The reference field must not be greater than 50 characters.',
+    },
+    {
+      label: 'Subject',
+      value: testData.repeat(26),
+      errorMessage: 'The subject field must not be greater than 100 characters.',
+    },
+  ];
+
+  it.each(fieldLengthDataSet)('should validate the field length', async ({ label, value, errorMessage }) => {
```
chrisrhymes, line 114: This and the next test below are using jest data sets. It's a way of passing in variables to use the same test script multiple times with different settings. 
More info here: https://dev.to/avantar/using-data-sets-in-your-jest-tests-2b1

src/library/slices/WebChat/WebChat.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+    fireEvent.click(startButton);
+
+    expect(component).toHaveTextContent('Subject');
+
+    const submitButton = getByRole('button', { name: 'Connect to webchat (opens in new window)' });
+
+    fireEvent.submit(submitButton);
+
+    await waitFor(() => {
```
chrisrhymes, line 78: This is using the [waitFor](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor) method to wait for the error messages to display after the submitButton is pressed.

mstanford-wnc (approved):
All good.  Please proceed

PR #499, opened 2024-06-27
Fix spacing issue with directory service list accordion

chrisrhymes:
After fixing the heading styles in the theme generator it applied additional margin to the accordion buttons as they utilise the H4 styles. This change overrides the heading margin back to zero. 

## Testing
- using the master branch, run `npm run dev` and check the Directory Service List stories to see the accordion items having additional margin
- Checkout this branch and run `npm run dev` and the Directory Service List accordions spacing should now be as expected.


DamienBurkeWN (approved):
Powerful stuff 👍

PR #513, opened 2024-07-25
Add toggle buttons

chrisrhymes:
Part of [Search engine improvements](https://github.com/FutureNorthants/northants-website/issues/1044).

Add toggle buttons for use in the search results page. This will indicate whether the search is filtered to service pages or news pages. 

The toggle buttons allow us to pass in an onClick event which we can use in the frontend to reload the search results with the news or services content type filter as required (as the page is server side rendered).

## Testing
- Check out this branch and run `npm install` then `npm run dev` to start storybook
- View the Components -> Toggle Buttons
- Test the styles toggling between the two buttons. 

## Future changes
- The Homepage can be refactored to make use of the toggle buttons to help remove duplicated code.

### DIFF
--- a/src/library/structure/SearchResultsList/SearchResultsList.styles.js
+++ b/src/library/structure/SearchResultsList/SearchResultsList.styles.js
@@ -89,7 +89,3 @@ export const SignpostContainer = styled.div`
     }
   }
 `;
-
-export const Published = styled.p`
-  font-weight: bold;
-`;
--- a/src/library/structure/SearchResultsList/SearchResultsList.styles.js
+++ b/src/library/structure/SearchResultsList/SearchResultsList.styles.js
@@ -89,3 +89,7 @@ export const SignpostContainer = styled.div`
     }
   }
 `;
+
+export const Published = styled.p`
+  font-weight: bold;
+`;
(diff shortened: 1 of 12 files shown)


DamienBurkeWN (approved):
👍

src/library/structure/SearchResultsList/SearchResultsList.tsx
```diff
@@ -29,6 +29,9 @@ const SearchResultsList: React.FunctionComponent<SearchResultsListProps> = ({
           <Styles.Result key={i}>
             {result.service && <Styles.ServiceArea>{result.service}</Styles.ServiceArea>}
             <Styles.Title href={result.link}>{result.title}</Styles.Title>
+            {result.published && result.service === 'News' && (
+              <Styles.Published>Published {result.published}</Styles.Published>
```
DamienBurkeWN, line 33: Is it worth having non-news articles display their last updated date here instead?

src/library/structure/SearchResultsList/SearchResultsList.styles.js
```diff
@@ -89,3 +89,7 @@ export const SignpostContainer = styled.div`
     }
   }
 `;
+
+export const Published = styled.p`
+  font-weight: bold;
```
DamienBurkeWN, line 94: My feeling is this is a bit overwhelming - perhaps a smaller font size and no bold unless the date is > 1 year?

DamienBurkeWN (commented):
See comments

src/library/structure/SearchResultsList/SearchResultsList.styles.js
chrisrhymes: Replaced with the NewsArticleOldBanner so it is consistent with the news article page.

src/library/structure/SearchResultsList/SearchResultsList.tsx
chrisrhymes: I'm now passing in a date so it is more flexible for future use.

DamienBurkeWN (approved):
👍

PR #521, opened 2024-08-09
Set secondary button background to white

### LINKED ISSUE #502: HeroImage call to action secondary contrast
When the HeroImage call to action is set to secondary and background box is set to true, the text contrast does not meet the AAA standard. Set the button background to white to meet the AAA contrast ratio. 

Box:
background-color: rgba(245, 245, 245, 0.95);

Button:
color: rgb(56, 88, 137);
background: transparent;

https://index.silktide.com/website/west-northamptonshire/july-2024/recommendations/aaa-text-contrast

chrisrhymes:
Resolves #502 

The HeroImage component uses the CallToAction slice, which uses the Button component... 

The background was set to transparent, but this meant that the contrast between the text and the background could be unknown. This updates the secondary button styles so that the background is white, ensuring accessible contrast between the blue text and the white background. 

## Testing
- View the Components -> Buttons -> Secondary and check the button has a white background 
- View the Structure -> Hero Image -> Hero Image With Custom Search and check the button has a white background
- Use the [Silktide browser plugin](https://silktide.com/toolbar/) to test the colour contrast between the button text and the button background.

### DIFF
--- a/src/library/components/Button/Button.styles.js
+++ b/src/library/components/Button/Button.styles.js
@@ -45,7 +45,7 @@ export const StyledButton = styled.a`
   &.button--secondary {
     color: ${(props) =>
       props.$colourOverride ? props.$colourOverride : props.theme.theme_vars.colours.action} !important;
-    background-color: transparent;
+    background-color: ${(props) => props.theme.theme_vars.colours.white};
     border: 3px solid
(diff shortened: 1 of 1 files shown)


DamienBurkeWN (approved):
👍

PR #530, opened 2024-09-05
1050 news releases are listed by month

chrisrhymes:
Add month and year as possible news filters. The drop down select and their options are passed in from the frontend. This change allows the filters to be used by different filters instead of hard coding it to only work with services. 

You can't really test this much on it's own. It depends on the frontend to work with the news filters. There should probably be a rework for this component so it just manages display and the frontend controls the onChange events. 

## Testing
- Checkout this branch `git fetch` then `git checkout 1050-news-releases-are-listed-by-month` then run `npm install` then `npm run dev`
- Manually test the DropDownFilter still works as expected


src/library/components/DropDownFilter/DropDownFilter.styles.js
```diff
@@ -1,6 +1,5 @@
-import styled, {css} from "styled-components";
-
-export const Container = styled.div`  
-  ${props => props.theme.fontStyles};
-`
+import styled, { css } from 'styled-components';
 
+export const Container = styled.div`
```
chrisrhymes, line 3: Added margin to separate the two year and month filters when hasMargin is true. The props are passed in when the component is used in the frontend.

src/library/structure/NewsArticleFilterAccordion/NewsArticleFilterAccordionText.js
```diff
@@ -1,20 +1,28 @@
...
+  },
+  services: {
+    title: 'Services',
+    queryParamKey: 'services',
+  },
+  articleType: {
+    title: 'Type of Article',
+    queryParamKey: 'articleType',
+  },
+  year: {
```
chrisrhymes, line 14: Main change is the year and month added. The rest is auto formatting by prettier.

src/library/components/DropDownFilter/DropDownFilter.tsx
```diff
@@ -1,26 +1,38 @@
...
+import { handleParams } from './../../helpers/url-helpers.js';
+import DropDownSelect from '../DropDownSelect/DropDownSelect';
+
+const DropDownFilter: React.FunctionComponent<DropDownFilterProps> = ({
+  id,
+  label,
+  options,
+  selected,
+  hideLabel = false,
+  filterKey = 'services',
```
chrisrhymes, line 14: Lots of prettier auto formatting in this file. The filterKey has been added. This allows the same component to be re-used by services, year and month filters. 

The optional hasMargin prop has also been added to allow some margin-bottom if two filters are on the page together.

mstanford-wnc (approved):
Component appropriately configurable.  Approving this

PR #533, opened 2024-10-18
Add promo tabs component

chrisrhymes:
A new component as part of the new homepage design. https://github.com/FutureNorthants/northants-website/issues/1102

This component allows 4 promo items to be in a tabbed structure to take up less space, but highlight more articles. 

Preview link available here: https://deploy-preview-533--northants-design-system.netlify.app/?path=/story/library-structure-promo-tabs--example-promo-tabs

## Testing
- Go to Structure -> Promo Tabs -> Example Promo Tabs and test styles all look ok in different browsers and that the tabs functionality works as expected. 
- Test in different screen sizes to see how it works on mobile, tablet and desktop
- Manually run tests with `npm run test`


src/library/components/Column/Column.test.tsx
```diff
@@ -96,4 +96,19 @@ describe('Column', () => {
     expect(listItem).toBeVisible();
     expect(listItem).toHaveTextContent('example content');
   });
+
+  it('should render the custom class, role and id', () => {
```
chrisrhymes, line 100: Added ability to set an optional role and id attribute to a column. This is to meet the ARIA tab role requirements. 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role

src/library/structure/PromoBlock/PromoBlock.storydata.ts
chrisrhymes, line 1: Updated the images to use lorem picsum as the placeholder.com seems to struggle to load images consistently.

src/library/structure/PromoTabs/PromoTabs.tsx
```diff
@@ -0,0 +1,85 @@
...
+    <Styles.Container data-testid="PromoTabs">
+      <Row>
+        <Column small="full" medium="one-third" large="one-quarter" classes="promo-tab-column">
+          <div role="tablist">
+            {promos.map((promo, index) => (
+              <Styles.TitleButton
+                onClick={() => toggleTab(index)}
+                key={index}
+                $isActive={index === activeTab}
+                role="tab"
```
chrisrhymes, line 27: Following the ARIA tab role specification. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role

src/library/structure/PromoTabs/PromoTabs.types.ts
chrisrhymes, line 1: Identical props as the Promo Banner and Promo Block. This makes them easily interchangeable on the frontend.

mstanford-wnc:
First impression - do we perhaps want a less prominent font, font combination or smaller 'font size' for links on the right hand side? (such as the one highlighted below)

At the moment, 'Promo Tabs' (highlighted on green, on the left) and hyperlinks (such as the one highlighted on yellow) utilise the same font:

image

When the website is viewed on a smartphone, this seems to be especially unintuitive as it breaks the hierarchy

image

Could the section names not currently being viewed (i.e. those highlighted in blue) be indented?

image

And/or made less prominent by temporarily making their font size smaller, as below?

image

chrisrhymes:
> First impression - do we perhaps want a less prominent font, font combination or smaller 'font size' for links on the right hand side? (such as the one highlighted below)
> 
> At the moment, 'Promo Tabs' (highlighted on green, on the left) and hyperlinks (such as the one highlighted on yellow) utilise the same font:
> 

@mstanford-wnc We use a single font throughout the site for consistency and in order to meet [WCAG minimum target size (AAA)](https://www.w3.org/WAI/WCAG22/Understanding/target-size-enhanced) we have to ensure links are a minimum of 44px high. 

You are right about the mobile layout. It doesn't work the way it is. I think the original idea was to replace the tabs with the PromoBlock component on mobile so they stacked vertically one above each other. I will update this PR to do this instead.

chrisrhymes:
@mstanford-wnc I have updated this PR so that it shows the PromoBlock on mobile and PromoTabs on larger screen sizes. This should fix any potential issues with the PromoTabs on mobile.

mstanford-wnc (approved):
Looks good to me
Approving this

PR #535, opened 2024-10-22
Add max length to webchat form and character count

### LINKED ISSUE #534: Textarea should have a maxlength
Feedback from the web chat testing has indicated that they would like a hard limit in the textarea to prevent users from typing more than the allowed amount of characters. 

This could be achieved with a maxlength attribute, but could also be enhanced with some kind of character count like this component in the Gov UK design system: https://design-system.service.gov.uk/components/character-count/

chrisrhymes:
Resolves #534 

- Updated the web chat form so the inputs and textarea have max length attribute set
- Add a character count component that states how many characters are remaining
- Added character count to textarea component when the max length is set

## Testing
- Checkout this branch `git fetch` then `git checkout 534-textarea-should-have-a-maxlength`
- Go to Pages -> Example Contact Page and click the Launch webchat button
- Test that the maxlength attribute is set on the inputs and the textarea
- Test that the character count counts down as you type on the subject textarea field

### DIFF
--- a/src/library/components/CharacterCount/CharacterCount.tsx
+++ b/src/library/components/CharacterCount/CharacterCount.tsx
@@ -1,0 +1,20 @@
+import React, { useEffect, useState } from 'react';
+import { CharacterCountProps } from './CharacterCount.types';
+import * as Styles from './CharacterCount.styles';
+
+const CharacterCount: React.FunctionComponent<CharacterCountProps> = ({ maxLength, input }) => {
+  const [remaining, setRemaining] = useState(maxLength);
+
+  useEffect(() => {
+    const calculatedRemaining = maxLength - input.length;
+    setRemaining(calculatedRemaining < 0 ? 0 : calculatedRemaining);
+  }, [input]);
+
+  return (
+    <Styles.Container data-testid="CharacterCount">
+      <Styles.Count aria-live="polite">You have {remaining} characters remaining.</Styles.Count>
+    </Styles.Container>
+  );
+};
+
+export default CharacterCount;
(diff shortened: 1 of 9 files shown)


src/library/components/CharacterCount/CharacterCount.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const [remaining, setRemaining] = useState(maxLength);
+
+  useEffect(() => {
+    const calculatedRemaining = maxLength - input.length;
+    setRemaining(calculatedRemaining < 0 ? 0 : calculatedRemaining);
+  }, [input]);
+
+  return (
+    <Styles.Container data-testid="CharacterCount">
+      <Styles.Count aria-live="polite">You have {remaining} characters remaining.</Styles.Count>
```
chrisrhymes, line 15: More info on aria-live="polite" here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

It should inform the user via screen reader that is has updated, but not so rapid as to be annoying.

src/library/components/Textarea/Textarea.styles.js
```diff
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 export const StyledTextarea = styled.textarea`
   ${(props) => props.theme.fontStyles}
   margin-top: 0 !important;
-  margin-bottom: 25px;
+  margin-bottom: ${(props) => (props.maxLength ? '0' : '25px')};
```
chrisrhymes, line 6: Margin is applied to the CharacterCount component when maxLength is set, so removing the margin bottom from the textarea itself.

mstanford-wnc:
As per the screenshots below, the count-down looks good to me:

image
image
image
image

The character limit on the email field should probably be a lot less than 100 though:  
image

PR #539, opened 2024-12-10
Include rate this page form content in html so it is translated

chrisrhymes:
- Include the form content in the HTML but use CSS to show and hide the relevant sections.
- Set the required fields only when showFullForm is true

## Testing
- Checkout this branch with `git fetch && git checkout 1178-rate-this-page-feature---doesnt-translate-fully-when-google-translate-widget-is-in-place`
- View Structure -> Rate This Page -> Example Rate This Page 
- Test the form content displays as expected and that the validation works as expected. 
    - If you select Yes then the other form fields should not be required. 
    - If you select No then it should display additional question boxes. Click 'I have feedback about the information on this page' then the rest of the form should display. 
- View Example Pages -> Content Page -> Example Content Page
    - Translate the page into another language from the header
    - Retest the above test and check that the Rate This Page form content and questions are now translated

## Frontend testing
- In the design system, run `yalc publish`
- In the frontend, run `yalc add northants-design-system` then `npm install` then `npm run dev`
- Visit a service page and retest the Rate This Page form behaves as expected.
- Translate the page using the header and then retest that the form content is also translated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9423,9 +9423,9 @@
       }
     },
     "node_modules/express": {
-      "version": "4.21.1",
-      "resolved": "https://registry.npmjs.org/express/-/express-4.21.1.tgz",
-      "integrity": "sha512-YSFlK1Ee0/GC8QaO91tHcDxJiE/X4FbpAyQWkxAvG6AXCuR65YzK8ua6D9hvi/TzUfZMpc+BwuM1IPw8fmQBiQ==",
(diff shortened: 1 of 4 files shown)


mstanford-wnc (approved):
Approving this.  The remaining items are now being translated as required:
image
image
This differs to how it was previously, where the content under the Yes/No radio buttons does not get translated:
image
image

PR #546, opened 2025-01-06
Accessibility fixes for google translate link

chrisrhymes:
Resolves https://github.com/FutureNorthants/northants-website/issues/1188
Resolves https://github.com/FutureNorthants/northants-website/issues/1187

- Add '(opens in new window)' text to link 
- Add padding around link to make it over 45px high (AAA rating)

## Testing
- Checkout this branch with `git fetch && git checkout 1188-accessibility---add-and-opens-in-new-link-text-next-to-power-by-google-translate-text`
- View Page examples -> Homepage -> Example homepage no search
- Expand the Translate menu in the header
- The link should now have `(opens in new window)` as part of the link text
- The link should have padding above and below
- The link should now have a focus state to indicate when the link is focused with the keyboard, as well as hover and active states

### DIFF
--- a/src/library/components/GoogleTranslate/GoogleTranslate.styles.js
+++ b/src/library/components/GoogleTranslate/GoogleTranslate.styles.js
@@ -26,7 +26,27 @@ export const Container = styled.div`
   }
 
   #google_translate_element {
-    margin-bottom: ${(props) => props.theme.theme_vars.spacingSizes.medium};
+    margin-bottom: ${(props) => props.theme.theme_vars.spacingSizes.small};
+
+    a {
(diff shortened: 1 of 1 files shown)


mstanford-wnc:
Confirmed that the padding and text has been appropriately extended:

image

I suspect it would be better if the 'opens in new window' text were to be replaced with an Icon, though:

image

chrisrhymes:
Asked web team for feedback in todays stand up meeting on whether to use the icon or text.

mstanford-wnc (approved):
Approving this.  All good.  Icon ruled-out

PR #555, opened 2025-01-21
Extended acceptable URLs to include one.network, and included speech mark fix

mstanford-wnc:
Extended list of acceptable URLs to include "one.network" but "one.network" refused to embed.  Pushing progress so far to GitHub.
Also included replace fix: `src_matches[0].replace('src="', '').replace('"', '')`


mstanford-wnc:
@chrisrhymes - Please can you review my code?

src/library/slices/GoogleMap/GoogleMap.tsx
```diff
@@ -20,20 +20,23 @@ const GoogleMap: React.FunctionComponent<GoogleMapProps> = ({
...
+
+    embed_url = acceptable_embedded_url_matches?.length == 1 ? embed_url : '';
   }
 
-  /* We also check the non-embed link goes to goo.gl/maps or www.google.com/maps */
+  /* We also check the non-embed link goes to goo.gl/maps or www.google.com/maps or one.network*/
   if (link_url) {
-    const googl_matches = link_url.match(
-      /^https:\/\/goo.gl\/maps|https:\/\/www.google.com\/maps|https:\/\/maps.app.goo.gl/gi
+    const acceptable_non_embedded_url_matches = link_url.match(
```
chrisrhymes, line 36: I have just asked for confirmation that the one.network page is ok as an alternative link when cookies are not enabled.

chrisrhymes (changes requested):
Tested and it works as expected, but just a few comments in the code. I'll let you know when I hear back about the one.network link.

src/library/slices/GoogleMap/GoogleMap.storydata.ts
```diff
@@ -50,3 +59,16 @@ export const GoogleMapWithUpdatedLinkFormat = {
...
+
+export const OneNetworkMapLink = {
+  title: OneNetworkCommon.title,
+  link_title: OneNetworkCommon.link_title,
+  link_url: OneNetworkCommon.link_url,
+  iframe_html: OneNetworkCommon.iframe_html,
+  allowCookies: true,
+};
+
+export const OneNetworkMapIncorrectOrgDetailsLink = { ...OneNetworkMapLink };
```
chrisrhymes, line 71: You can spread another object within the same object, rather than overwriting individual object keys. 

Also, this no longer shows the Invalid map alert message as the component no longer checks for the organisation id. I have amended the URL in the below suggestion to `apis-gb`, rather than `api-gb` so it fails. You could even make it more explicit by changing it to something like `incorrect-url.one.network` 

```suggestion
export const OneNetworkMapIncorrectOrgDetailsLink = {
  ...OneNetworkMapLink,
  ...{
    title: 'One Network Map (Incorrect Organisation IDs)',
    iframe_html:
      '<iframe src="https://apis-gb.one.network/embedded/?options=%7B%22organisationID%22%3A2248" title="Causeway one.network map" style="width: 800px; height: 559px; border: none;"></iframe>',
  },
};
```

src/library/slices/GoogleMap/GoogleMap.tsx
chrisrhymes: David has replied confirming the one.network page is fine for the link_url. 👍

chrisrhymes (approved):
All good 👍

PR #559, opened 2025-01-30
Check inquest appointment type includes the hearing term

chrisrhymes:
Resolves #1204

Check if the appointment type contains the relevant term to group the appointments. It was unknown that the appointment type contained the initials of the coroner. 

## Testing
- Checkout this branch `git fetch && git checkout 1204-inquest-schedule-slice-hearings-filter-needs-to-be-updated`
- View the Slices -> Inquest Schedule and check that the items are still grouped correctly with the updated appointment types in the story data file


mstanford-wnc:
Component works as required.

`npm run test` was successful:

image

mstanford-wnc (approved):
Component works as required.

`npm run test` was successful

PR #564, opened 2025-02-10
Tweaked Accordion Styles to include Padding and Margin changes

mstanford-wnc:
Resolves: https://github.com/FutureNorthants/northants-website/issues/1203

### DIFF
--- a/src/library/slices/Accordion/Accordion.styles.js
+++ b/src/library/slices/Accordion/Accordion.styles.js
@@ -47,6 +47,7 @@ export const VisuallyHidden = styled.span`
 
 export const Section = styled.div`
   padding-top: 0;
+  padding-left: 0px;
 `;
 
 export const SectionHeader = styled.div`
@@ -68,22 +69,24 @@ export const SectionHeader = styled.div`
 export const SectionHeading = styled.div`
   margin-top: 0 !important;
   margin-bottom: 0 !important;
+  padding-bottom: 0 !important;
+  padding-top: 0 !important;
 `;
 
 const SectionButtonIsFilteredStyles = (props) => {
   if (props.$isFilter) {
     return css`
       ${(props) => props.theme.theme_vars.h4}
-      padding-top: 15px;
-      padding-bottom: 15px;
+      padding-top: 10px;
+      padding-bottom: 10px;
       margin-top: 0;
       margin-bottom: 0;
     `;
   } else {
     return css`
       ${(props) => props.theme.theme_vars.h3}
-      padding-top: ${(props) => props.theme.theme_vars.spacingSizes.small};
-      padding-bottom: ${(props) => props.theme.theme_vars.spacingSizes.small};
+      padding-top: ${(props) => props.theme.theme_vars.spacingSizes.extra_small};
+      padding-bottom: ${(props) => props.theme.theme_vars.spacingSizes.extra_small};
       margin-top: 0;
       margin-bottom: 0;
     `;
@@ -93,9 +96,10 @@ const SectionButtonIsFilteredStyles = (props) => {
 export const SectionButton = styled.button`
   display: inline-block;
   margin-top: 0;
-  margin-bottom: 0;
+  margin-bottom: 70;
   margin-left: 0;
-  padding: 0;
+  padding: 0 ${(props) => props.theme.theme_vars.spacingSizes.extra_small} 0
+    ${(props) => props.theme.theme_vars.spacingSizes.extra_small};
   border-width: 0;
   color: inherit;
   background: none;
@@ -149,7 +153,10 @@ export const ReadMore = styled.button`
   border: none;
   margin-top: 15px;
   margin-bottom: 10px;
-  padding: 0;
+  margin-left: ${(props) => props.theme.theme_vars.spacingSizes.extra_small};
+  //padding: 0 0 0 ${(props) => props.theme.theme_vars.spacingSizes.small};
+  margin-left: 15px;
+  padding: 0px;
   font-size: ${(props) => props.theme.theme_vars.fontSizes.extra_small};
 `;
 export const ShowLessButton = styled.button`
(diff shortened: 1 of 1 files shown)


mstanford-wnc:
Padding and indent added as required.  Screenshots to follow

src/library/slices/Accordion/Accordion.styles.js
```diff
@@ -68,22 +69,24 @@ export const SectionHeader = styled.div`
...
+  padding-top: 0 !important;
 `;
 
 const SectionButtonIsFilteredStyles = (props) => {
   if (props.$isFilter) {
     return css`
       ${(props) => props.theme.theme_vars.h4}
-      padding-top: 15px;
-      padding-bottom: 15px;
+      padding-top: 10px;
```
chrisrhymes, line 80: These reductions in padding mean that the accordion buttons now fail the minimum target size check [WCAG 2.1 AAA 2.5.5](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html).

src/library/slices/Accordion/Accordion.styles.js
```diff
@@ -68,22 +69,24 @@ export const SectionHeader = styled.div`
...
       margin-top: 0;
       margin-bottom: 0;
     `;
   } else {
     return css`
       ${(props) => props.theme.theme_vars.h3}
-      padding-top: ${(props) => props.theme.theme_vars.spacingSizes.small};
-      padding-bottom: ${(props) => props.theme.theme_vars.spacingSizes.small};
+      padding-top: ${(props) => props.theme.theme_vars.spacingSizes.extra_small};
+      padding-bottom: ${(props) => props.theme.theme_vars.spacingSizes.extra_small};
```
chrisrhymes, line 89: Same as previous comment about minimum target size.

src/library/slices/Accordion/Accordion.styles.js
```diff
@@ -93,9 +96,10 @@ const SectionButtonIsFilteredStyles = (props) => {
 export const SectionButton = styled.button`
   display: inline-block;
   margin-top: 0;
-  margin-bottom: 0;
+  margin-bottom: 70;
```
chrisrhymes, line 99: This is an invalid property value as it doesn't have a unit, such as `px`. Not sure that 70px is needed here?

src/library/slices/Accordion/Accordion.styles.js
```diff
@@ -93,9 +96,10 @@ const SectionButtonIsFilteredStyles = (props) => {
...
   margin-left: 0;
-  padding: 0;
+  padding: 0 ${(props) => props.theme.theme_vars.spacingSizes.extra_small} 0
```
chrisrhymes, line 101: The padding setting here is overwritten by the SectionButtonIsFilteredStyles function below. Might cause confusion having it in two places..

src/library/slices/Accordion/Accordion.styles.js
```diff
@@ -149,15 +153,19 @@ export const ReadMore = styled.button`
   border: none;
   margin-top: 15px;
   margin-bottom: 10px;
-  padding: 0;
+  margin-left: ${(props) => props.theme.theme_vars.spacingSizes.extra_small};
+  //padding: 0 0 0 ${(props) => props.theme.theme_vars.spacingSizes.small};
```
chrisrhymes, line 157: Please remove this commented line if it's not needed.

src/library/slices/Accordion/Accordion.styles.js
```diff
@@ -149,15 +153,19 @@ export const ReadMore = styled.button`
...
+  margin-left: 15px;
```
chrisrhymes, line 158: You have margin-left here twice in the same declaration. Please remove one that's not needed.

src/library/slices/Accordion/Accordion.styles.js
```diff
@@ -174,10 +182,13 @@ export const ShowLessButton = styled.button`
 
 export const AccordionIcon = styled.span`
   position: absolute;
-  top: 22px;
+  top: 15px;
   right: 4px;
   width: 34px;
   height: 32px;
+  margin-top: 0px !important;
```
chrisrhymes, line 189: Consider if !important is really needed here. Does it work without it?

src/library/slices/Accordion/Accordion.styles.js
```diff
@@ -189,29 +200,34 @@ export const AccordionIcon = styled.span`
...
       transform: rotate(-45deg);
-      top: 10px;
+      top: 5px;
     }
   }
 `;
 
 export const SectionSummary = styled.div`
-  margin-top: 10px;
+margin-left: 0px
```
chrisrhymes, line 217: Please check if you are using prettier as the alignment seems off here. Also remove commented out line below if not needed.

src/library/slices/Accordion/Accordion.styles.js
```diff
@@ -189,29 +200,34 @@ export const AccordionIcon = styled.span`
...
   color: ${(props) => props.theme.theme_vars.colours.black};
-`;
+  `;
 
 export const SectionContent = styled.div`
   display: none;
-  padding-top: 5px;
+  padding-top: 15px;
   padding-bottom: 15px;
+  // padding-left: 25px;
```
chrisrhymes, line 229: Commented out line.

src/library/slices/Accordion/Accordion.styles.js
```diff
@@ -222,6 +238,11 @@ export const SectionContent = styled.div`
     }
   }
 
+  p {
+    padding-left: 3px;
+    margin-left: 0px;
+  }
+
```
chrisrhymes, line 245: Consider that the accordion section content may have different html in here too, for example such as a div, ol or ul, which would make the paragraph not line up with the other content. Better to add padding to the parent and let the child elements align with that, otherwise there will be lots of declarations for different cases here.

chrisrhymes (changes requested):
Added a few comments in line. 
- Please consider if !important is really needed and only use where necessary. 
- Try to remove commented out css lines as leaving them in creates a lot of noise/confusion later
- Be careful about declaring the same property multiple times in the same declaration, such as `margin-left` in one example. 
- Think about if you can add a style to a parent element and let it cascade down, rather than applying specific styles to child elements.

PR #568, opened 2025-02-11
Updated Global Style Reset TSX to include anchor-tag link-item styling

mstanford-wnc:
Resolves:  https://github.com/FutureNorthants/northants-website/issues/1081

### DIFF
--- a/src/themes/GlobalStyleReset.tsx
+++ b/src/themes/GlobalStyleReset.tsx
@@ -133,10 +133,12 @@ export const GlobalStyleReset: any = createGlobalStyle<any>`
     left: 1.25rem;
     padding-right: 1.25rem;
   }
-  li a {
-    padding-bottom: 11px;
-    padding-top: 12px;
+
+  li > a {
+    display: inline-block;
+    min-height: 44px;
   }
+
   blockquote,
   q {
     quotes: none;
--- a/src/themes/GlobalStyleReset.tsx
+++ b/src/themes/GlobalStyleReset.tsx
@@ -133,12 +133,10 @@ export const GlobalStyleReset: any = createGlobalStyle<any>`
     left: 1.25rem;
     padding-right: 1.25rem;
   }
-
   main > ul > li > a {
     display: inline-block;
     min-height: 44px;
   }
-
   blockquote,
   q {
     quotes: none;
--- a/src/themes/GlobalStyleReset.tsx
+++ b/src/themes/GlobalStyleReset.tsx
@@ -133,6 +133,10 @@ export const GlobalStyleReset: any = createGlobalStyle<any>`
     left: 1.25rem;
     padding-right: 1.25rem;
   }
+  li a {
+    padding-bottom: 11px;
+    padding-top: 12px;
+  }
   blockquote,
   q {
     quotes: none;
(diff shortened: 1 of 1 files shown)


mstanford-wnc:
@chrisrhymes - Please can you test this?

### Testing Prep

+ Checkout this branch via: `git checkout fix/silktide-issue-anchor-tag-size-in-long-ul`
+ Launch the design system via: `npm run dev`
+ Launch an instance of Google Chrome that has the silk-tide accessibility plug-in

### Testing the current design config

+ Navigate to `https://design-system.westnorthants.gov.uk/`
+ Follow Testing steps 1-4 below
+ Confirm that, as per the screenshot immediately below, there are 6 "Large enough controls" issues, including 3 that reference the **'list link'** anchor tags

image

### Testing the branched config

+ In another Google Chrome tab, navigate to: `http://localhost:6006/`
+ Repeat steps 1-4 below
+ Confirm that, as per the screenshot immediately below, there are only now **3** "Large enough controls' issues (i.e. the Anchor Tag issues are no longer listed)
+ Scroll-down on the right hand pane until you see the **list link** anchor tags and confirm that they look as required.

image

### Testing Steps

1.  Expand: `Page Examples` -> `Content Page` -> `Example Content Page`
2.  Click on the 'new window' icon pictured below

image

3. Launch the 'Silktide Accessibility Checker' as below:

image

4. Choose 'Accessibility Checker' and 'WCAG 2.2 AAA' from the drop-down:

image

src/themes/GlobalStyleReset.tsx
```diff
@@ -133,6 +133,10 @@ export const GlobalStyleReset: any = createGlobalStyle<any>`
     left: 1.25rem;
     padding-right: 1.25rem;
   }
+  li a {
```
chrisrhymes, line 136: Although this resolves the automated accessibility checks, it may cause a usability issue. The links now overlap each other so there is potential for clicking on either the link above or below when trying to click on `list link 2` in this example. 

Screenshot 2025-02-12 at 08 43 57

Perhaps try setting the display to inline-block? 

Also, you could make this a little more specific so it targets anchor tags that are direct descendants of li, instead of all anchor tags in li's by using the [CSS child combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator). This way if an anchor tag is inside a paragraph tag that's inside an li it won't be affected.

chrisrhymes (changes requested):
Added some comments inline.

mstanford-wnc:
Thanks for your feedback @chrisrhymes.  Please can you re-test ?

The hyperlinks have now been spaced-out a bit, such that, as you can see below, they are no longer overlapping:

image

This unfortunately means, that, as per the screenshot below, the gaps are a bit bigger than I'd like:

image

src/themes/GlobalStyleReset.tsx
```diff
@@ -133,6 +133,12 @@ export const GlobalStyleReset: any = createGlobalStyle<any>`
     left: 1.25rem;
     padding-right: 1.25rem;
   }
+
+  li > a {
```
chrisrhymes, line 137: I was thinking about what you said earlier and I like your idea to make this specific to the list and links only in the `<main>`. This means it won't affect the header and footer. 

Please can you change `li > a` to `main li > a`?

mstanford-wnc: Thanks @chrisrhymes .  That's now changed.
Please can you re-test?

chrisrhymes (approved):
Nice one! Tested this change out locally on various pages that previously reported issues in the [silktide portal](https://app.eu.silktide.com/westnorth/manager/websites/2/0/wcag-21-checks/checks/ensure-reasonable-target-size?tabId=pages) and the issue seems to be resolved. 👍

PR #580, opened 2025-03-11
Adjust line height for list items with links

chrisrhymes:
Resolves https://github.com/FutureNorthants/northants-website/issues/1232

Increase the line height of li to 44px so it is the same as the link. 

## Testing
- Run `npm run dev`
- Visit the Page Examples -> Content page, then scroll down to the lists and the lists with links to preview the changes. 
- The list items should now have consistent spacing whether there are links or not.


rupertfoleywnc (approved):
Line heights for List Items are now consistent with or without an Anchor tag on the line.

PR #583, opened 2025-03-13
Fix line height issue with section links

chrisrhymes:
Resolves https://github.com/FutureNorthants/northants-website/issues/1081

After applying the fix for list items line height, this has made the section links spacing look wrong (as per screenshot below).
Screenshot 2025-03-13 at 11 01 33

This PR overwrites the default line height for the section links to make them look more like they did previously. 

## Testing
- Checkout this branch with `git fetch && git checkout fix/section-links-line-height`
- Visit the Example pages -> Service Landing Pages` and view the different examples, ensuring that the line spacing of the section links now looks more like the live website.

### DIFF
--- a/src/library/structure/SectionLinks/SectionLinks.styles.js
+++ b/src/library/structure/SectionLinks/SectionLinks.styles.js
@@ -1,7 +1,11 @@
 import styled from 'styled-components';
 import Heading from '../../components/Heading/Heading';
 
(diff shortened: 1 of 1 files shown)


rupertfoleywnc (approved):
Compared example page to wnc website and spacing looks to be the same at both small and large widths.

PR #591, opened 2025-04-03
Add Cludo Search component

chrisrhymes:
Resolves https://github.com/FutureNorthants/northants-website/issues/1183

Use the Cludo Search components to implement Cludo Search. The results are one component and the search bar used in the header is another. 

**We are currently using a pre-release version that supports server-side rendering for Next.js**

## Testing
- In your `.env` file (create if you don't have one), set the `NEXT_PUBLIC_CLUDO_CUSTOMER_ID=` and `NEXT_PUBLIC_CLUDO_ENGINE_ID=` (ask for the values).
- View the Structure -> Cludo Search and test the search results. Search for a search term and test the results return as expected and the layout looks ok. Ensure the autocomplete appears when typing, such as 'bins' and use the mouse and keyboard to navigate and select an autocomplete result.
- View the Structure -> Cludo Search Bar and test the autocomplete displays as expected.


mstanford-wnc (approved):
Approved

PR #618, opened 2025-06-09
Update image ratio on event link

chrisrhymes:
- Fix the image ratio for the event link image. It should be 16by9 ratio to match the image size in the CMS. 
- Add event type checkboxes. This also has an update for CheckboxListFilter to make it more reusable, instead of purely hardcoded for the news page.
- Add cards for links to other events pages

## Testing
- Checkout this branch with `git fetch && git checkout fix/event-link-image-ratio`
- Go to Structure -> EventList -> Example Event List 
- Test that the cards now show, the event image is 16by9 ratio and that the event type filters are now showing


mstanford-wnc (approved):
Approving this.  All looking good.

image

PR #627, opened 2025-06-25
Bugfix/autocomplete search results

mstanford-wnc:
## Testing
- checkout this branch
- run: `yalc publish`
- make sure the `docker proxy` is running
- checkout the back-end branch: `preprod_backend`
- on the back-end branch, execute:
    - `git pull`
    - `make` 
- checkout the front-end branch: `preprod_frontend`
- on the front-end branch, execute: 
    - `git pull` 
    - `npm install`
    - `yalc add northants-design-system`
    - `npm run dev`
- navigate to `http://localhost:3000` or the equivalent
- shrink the width of the window so that the search box is just above the 'Pay' and other Hero links
- Do a search for 'P' and confirm that the search suggestions are positioned above the 'Hero links':

### DIFF
--- a/src/library/components/Autocomplete/Autocomplete.styles.js
+++ b/src/library/components/Autocomplete/Autocomplete.styles.js
@@ -38,7 +38,7 @@ export const AutocompleteLabel = styled.label`
 export const AutocompleteSuggestionList = styled.ul`
   position: absolute;
   display: block;
-  z-index: 1;
+  z-index: 100;
   width: auto;
   font-family: ${(props) => props.theme.theme_vars.fontstack};
   border-radius: ${(props) =>
--- a/src/library/structure/ServicesLinksBoxed/ServicesLinksBoxed.styles.js
+++ b/src/library/structure/ServicesLinksBoxed/ServicesLinksBoxed.styles.js
@@ -72,7 +72,7 @@ export const ServiceHeading = styled.span`
   line-height: 1.25;
   ${(props) => props.theme.theme_vars.h3}
   margin-top: 0;
-  margin-bottom:0;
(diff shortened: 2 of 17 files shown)


src/library/structure/ServicesLinksBoxed/ServicesLinksBoxed.styles.js
```diff
@@ -81,7 +81,7 @@ export const QuickLinksContainer = styled.div`
   overflow: hidden;
   top: auto;
   margin-top: ${(props) => props.theme.theme_vars.spacingSizes.medium};
-  z-index: 9;
+  z-index: 40;
```
chrisrhymes, line 84: This one needs to be lower than the back to top and quick exit button. On certain screen sizes the back to top button goes under the QuickLinksContainer. 
Screenshot 2025-07-01 at 09 04 11

src/library/components/Autocomplete/Autocomplete.styles.js
```diff
@@ -38,7 +38,7 @@ export const AutocompleteLabel = styled.label`
 export const AutocompleteSuggestionList = styled.ul`
   position: absolute;
   display: block;
-  z-index: 1;
+  z-index: 50;
```
chrisrhymes, line 41: Even though this has a higher z-index than the quick exit button, the quick exit button still shows above the autocomplete in the search bar. I'm not sure what should be higher in this specific situation though. 

Screenshot 2025-07-01 at 09 06 48

chrisrhymes (changes requested):
Generally ok, but just a couple of points to look into. I think we should discuss whether the quick exit button should be above or below the search auto complete in a stand up.

PR #657, opened 2025-08-27
Fix/add pre search suggestions to autocomplete

chrisrhymes:


chrisrhymes:
Provided some more information to Cludo and they will work on an update in the new year.

PR #674, opened 2025-09-24
Tweaked PostCodeSearch to include Waste

mstanford-wnc:
### Testing
- Checkout this branch
- run `npm run dev`
- Navigate to `LIBRARY -> Components -> Postcode Search`
- Test both the `Default Example` and `Waste Example` with the following post codes
   - NN4 6AT
   - NN5 6BB

### DIFF
--- a/src/library/components/PostCodeSearch/PostCodeSearch.stories.tsx
+++ b/src/library/components/PostCodeSearch/PostCodeSearch.stories.tsx
@@ -16,9 +16,21 @@ const meta: Meta<typeof PostCodeSearch> = {
 export default meta;
 type Story = StoryObj<typeof meta>;
 
-export const Example: Story = {
+export const DefaultExample: Story = {
   args: {
     otherCouncilLink: 'https://www.westnorthants.gov.uk',
+  },
+  render: (args) => (
+    <MaxWidthContainer>
+      <PostCodeSearch {...args} />
+    </MaxWidthContainer>
+  ),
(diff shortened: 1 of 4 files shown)


src/library/structure/SignpostLinks/SignpostLinks.stories.tsx
```diff
@@ -42,11 +43,9 @@ export const NorthSignpostLinksExample: Story = {
   },
   render: (args) => (
     <>
-      <p>
-        <i>
-          Make sure you have the right theme chosen for the corresponding example, otherwise this might not work right.
-        </i>
-      </p>
+      <Styles.RenderingMessage>
```
chrisrhymes, line 46: As these styles are only used in the story and not the component, we don't really want to export them with the rest of the styles when the design system is built and packaged up. You can make the styles inline here as the stories are not exported in the build.

```tsx
<p style={{ color: 'red', marginTop: 0, marginBottom: '12px' }}>
  Make sure you have the right theme chosen for the corresponding example, otherwise this might not work right.
</p>
```

src/library/structure/SignpostLinks/SignpostLinks.types.ts
chrisrhymes, line 1: Just for consistency and readability, please can you add a line space under each type to separate it from the one below.

chrisrhymes (approved):
All looks good!

